PR #47, opened 2021-11-03
Refactor and update ubiquity inventory component

chriswudev:


Zequez (changes requested):
Alright, so everything looks good so far.
Some improvements I would love to see are:
- Use the balance data from the global context instead of loading within this component; this way the component just displays the values but doesn't need to know how to load them.
- Extract each of the balances display into a separate component (you can make it on the same file) so we don't need to repeat https://gist.github.com/Zequez/6d5b14f0d528d06995071596a5505798

Zequez (approved):
Beautiful! Thank you! <3

Zequez:
Oh, I just realized, remember that for the Token component we need to be able to specify the number of decimals, since USDC has 6 decimals instead of the usual 18! I just made a quick fix with that change, so don't worry!

@pavlovcik this is quite ready for merging! The inventory components got quite a cleaning!

Zequez:
@pavlovcik any reason not to merge this?

0x4007:
Haven't gotten around to reviewing yet

PR #48, opened 2021-11-04
Global indicator

chriswudev:


Zequez (commented):
Looks good so far. There are still some of the components which also make transactions and weren't changed to use this global indicator. Do you intend to continue with those before merging?

chriswudev:
Yup, for sure!

Zequez (approved):

Zequez (approved):

Zequez:
Alright it all looks great! The only thing I would change is to make the indicator fixed positioned on the page, that way it's always visible even when the page is scrolled down; and we can set `pointer-events: none`, I already got the change locally so I'll commit it.

Zequez:
@pavlovcik this looks ready for merging

0x4007 (approved):

Issue #38, opened 2021-10-25
New DebtCoupons UI


Zequez:
So in accordance to the specs to mint uAR directly from the dashboard, I took the liberty to rework the existing Debt Coupons UI and integrate all the new upcoming features too. It's still a WIP and I'll be working on creating a bunch of sub-tasks to make the building process clearer and allow us to distribute the load with @chriswudev or anyone who wants to contribute.

<img width="585" alt="Captura de Tela 2021-10-25 às 09 30 06" src="https://user-images.githubusercontent.com/126037/138719085-9cdcbb46-6069-4f42-a18c-f2b24a795768.png">

## Sub-tasks

- https://github.com/ubiquity/uad-ui/issues/50
- https://github.com/ubiquity/uad-ui/issues/51
- https://github.com/ubiquity/uad-ui/issues/54
- https://github.com/ubiquity/uad-ui/issues/55
- https://github.com/ubiquity/uad-ui/issues/56
- https://github.com/ubiquity/uad-ui/issues/57
- https://github.com/ubiquity/uad-ui/issues/58
- https://github.com/ubiquity/uad-ui/issues/59
- https://github.com/ubiquity/uad-ui/issues/60

0x4007:
We should definitely explore some d3.js action when we have some time for custom charting/visualizations! Also I think `uDEBT` looks better than `uDebt` as a heads up

0x4007:
Also @Zequez I encourage you to work with April to set up bounties so we can have some new people from the community come contribute code. Generally speaking we can be very generous with `UBQ` (we print it out of thin air, more or less) then less generous with `uAR`, then `uAD` then lastly stables like `DAI/USDC/USDT`

Zequez:
> We should definitely explore some d3.js action when we have some time for custom charting/visualizations! Also I think `uDEBT` looks better than `uDebt` as a heads up

It would be interesting to see a chart of the TWAP price going up/down to see the debt cycle history.

0x4007:
So uDEBT doesn't have a depreciating rate in case it isn't clear. It holds its value until the moment it expires in six months. 

![image](https://user-images.githubusercontent.com/4975670/149848823-4f59665d-5533-4d94-96df-bf0704fe6515.png)

0x4007:
Can somebody explain the math to me here, doesn't seem right!

[...]

PR #52, opened 2021-11-18
New debt coupon UI

chriswudev:


Zequez:
Looking good so far! Keep it up! Thanks!

Zequez (commented):
We'll merge this when the whole #38 is finished;
@chriswudev remember to tag the issues on the PR description as you check them off, so they get linked back to this PR; and feel free to edit the issues and check the boxes!

components/common/contracts-shortcuts.ts
```diff
@@ -30,6 +32,8 @@ export async function accountBalances(account: EthAccount, contracts: Contracts)
     erc1155BalanceOf(account.address, contracts.debtCouponToken),
     erc1155BalanceOf(account.address, (contracts.bondingToken as unknown) as ERC1155Ubiquity),
     contracts.usdc.balanceOf(account.address),
+    erc1155BalanceOf(account.address, (contracts.uBOND as unknown) as ERC1155Ubiquity), // uBOND
```
Zequez, line 35: uBOND will be like uAR, it's an ERC20 token, so a simple balanceOf should suffice.

contracts.ts
```diff
@@ -213,6 +215,8 @@ export async function connectedContracts(): Promise<{
       crvToken: contracts.crvToken(addr.crvToken, provider),
       bondingToken: contracts.bondingToken(addr.bondingToken, provider),
       debtCouponToken: contracts.debtCouponToken(addr.debtCouponToken, provider),
+      uBOND: contracts.bondingToken(addr.bondingToken, provider),
```
Zequez, line 234: Confusingly, uBOND doesn't stand for the bondingToken 😆 
For now just put the uAR contract here, as the uBOND token doesn't exist yet.

contracts.ts
```diff
@@ -213,6 +215,8 @@ export async function connectedContracts(): Promise<{
...
+      uDEBT: contracts.debtCouponToken(addr.debtCouponToken, provider),
```
Zequez, line 235: We already have the debtCouponToken, why duplicate it?

components/common/contracts-shortcuts.ts
```diff
@@ -30,6 +32,8 @@ export async function accountBalances(account: EthAccount, contracts: Contracts)
...
+    erc1155BalanceOf(account.address, contracts.uDEBT), // uDEBT
```
Zequez, line 36: This returns the total balance as a BigNumber sum. But for the UI we'll need each token differentiated; check how the `erc1155BalanceOf` is created, and you'll see we first get a list of IDs for an address, and then we call the balanceOf of each token ID. We need the ID and the balance both, so we can then use it to sell these tokens individually.
Then we'll also need to expiration block, but we'll add a pin to that as we need to check how to get it from the DebtCoupon contract.

components/DebtCoupon.tsx
```diff
@@ -1,222 +1,521 @@
...
+  const udebtUbqRedemptionRate = 0.25;
+  const uadTotalSupply = 233000;
+  const ubondTotalSupply = 10000;
+  const uarTotalSupply = 30000;
+  const udebtTotalSupply = 12000;
+  const coupons: Coupons = {
+    uDEBT: [
+      { amount: 1000, expiration: 1640390400000, swap: { amount: 800, unit: "uAR" } },
+      { amount: 500, expiration: 1639526400000, swap: { amount: 125, unit: "uAR" } },
+      { amount: 666, expiration: 1636934400000, swap: { amount: 166.5, unit: "UBQ" } },
```
Zequez, line 139: Good idea the swap object for the mock!

contracts.ts
0x4007: Open to ideas for renaming the tokens (we have a rename function!)

0x4007:
@marcomow @Zequez should this be merged? Can you guys merge it if so?

sergfeldman:
Chris:
This pull request was implemented in another branch. 
The Master branch already has all the changes implemented in this pull request. 
That`s why there was no need to keep this pull request open.

PR #82, opened 2022-02-11
Enhancement/mono repo

0x4007:


contracts/dollar/.github/workflows/CI.yml
```diff
@@ -0,0 +1,40 @@
+name: CI
```
0x4007, line 1: should probably re set up CI

PR #125, opened 2022-05-03
Fix Vercel deployment

Zequez:
There were many issues, it wasn't just Vercel after all.
The `next build` command had some issues too.
The first thing was an issue with the paths on package.json; they need to be "./contracts/whatever" instead of "contracts/whatever" otherwise they don't seem to work.
Then there were some issues with conflicting versions of tools between the ubiquistick and dollar repositories; the typings generated were miss-matching and some things weren't working.
And there were some prettier errors too; I just turned them into warnings for the time being; I'll clean up later, they have no effect on the functionalities anyway.
There were some issues with the type matching of ERC20-based tokens that I ended up having to force a type casting. It seemed to be something about the newest version of Typechain, because it used to work with the older version.
And there was an issue in the way we were importing and re-exporting the PriceStabilization page that made Next.js freak out.
Over time we should work on the warnings, because the deployment is full of warnings lol

Vercel build command: `yarn build`
Vercel install command: `yarn install && yarn install-env && yarn install-contracts`

Preview https://ubiquity-dollar-development-three.vercel.app

Should close #99

PR #144, opened 2022-05-04
Develop

0x4007:
Because the purpose of this entire repository is to act as the "develop branch" perhaps we should delete the develop branch and just use main instead? What do you guys think makes the most sense? 

This PR I'm doing now is useless. The useful PR was #125


sergfeldman (approved):

sergfeldman:
I agree to use only the main branch.
The main point is to have a clear agreement with each developer who uses this repo.

sergfeldman:
@Zequez agrees to use only Main 
@ubq-cw Ok with this

Issue #89, opened 2022-04-18
Launch Party - Calculate APY by compounding


Zequez:
Right now the APY is calculated without compounding taken into account.

Right now it's calculated like this:

```ts
const apy = (reward.toNumber() / 1_000_000_000 / 5) * 365 * 100;
```

We gotta calculate it as if we compounded it every 5 days instead.

Zequez:
I'll share the math here, tag Christo and if everything checks out we'll close it.

Zequez:
Alright, so the math is very simple:

```ts
let multiplier = reward / 1_000_000_000
let apy = multiplier ** (365 / 5)
```

The reward is based on the contract reward that's stored as parts per billion. So a 1_100_000_000 reward multiplies the deposit by 1.1 over the course of 5 days. So to simulate compounding every five days for a year we just elevate it to (365/5). Is this math sound?

I'm thinking maybe we have to reduce the result by 1? (because we want to only show the APY of the REWARD, not the initial deposit that we give back, right?)

I mean, sure, it doesn't make much sense because you cannot actually compound the reward because it's on a different currency.

@hpiyankov could I get your input here?

hpiyankov:
1. Your multiplier ** (365 / 5) logic is correct as long as for a 10% multiplier you are using 1.1
2. Yes you need to subtract 1 at the end to get the APY

For the rest, I cannot comment, because it has been such a long time, that this is totally out of context for me.

PR #150, opened 2022-05-09
Fix APY calculation #89

Zequez:
Finishes the Issue #89 about the APY calculation, I  reduced the APY by 1 after calculating it.


0x4007 (approved):
Hristo approved so I'm sure its ok!

PR #204, opened 2022-07-11
Build static dapp 

zapaz:
Add build:dapp to build static website for IPFS hosting


zapaz:
works with fleek.co with command "yarn && yarn build:dapp" on nextjs env

PR #247, opened 2022-09-05
Fix/walletconnect

codeaddict99:
✔ Fixed wallet popup
✔ Fixed walletconnect qrcode modal


packages/dapp/components/lib/hooks/useWeb3.tsx
```diff
@@ -50,13 +60,54 @@ export const UseWeb3Provider: React.FC<ChildrenShim> = ({ children }) => {
...
+      // create new session
+      await connector.createSession();
+    }
+
+    connector.on("connect", async (error, payload) => {
+      const { chainId, accounts } = payload.params[0];
+      const newWalletAddress = accounts[0];
+
+      const newProvider = new WalletConnectProvider({
+        infuraId: process.env.API_KEY_ALCHEMY,
```
cuddle-dev: This might be wrong. Alchemy is a different service with infura.io. WalletConnectProvider supports 2 available options (rpc & infura). iirc, they don't support alchemy right now.

packages/dapp/next.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  publicRuntimeConfig: {
+    API_KEY_ALCHEMY: process.env.API_KEY_ALCHEMY,
+    API_KEY_COINMARKETCAP: process.env.API_KEY_COINMARKETCAP,
+    API_KEY_ETHERSCAN: process.env.API_KEY_ETHERSCAN,
+    MNEMONIC: process.env.MNEMONIC,
```
cuddle-dev: Actually these env variables are being used in contract packages.  I think we don't need configure  them(API_KEY_COINMARKETCAP, API_KEY_ETHERSCAN, MNEMONIC) in dapp package

cuddle-dev (changes requested):
Sometimes we might have rpc urls from other services (except infura.io). In this case, we will configure WalletConnectProvider using rpc urls. That would be better to make both (infura, rpc) configurable via .env.

packages/dapp/components/lib/hooks/useWeb3.tsx
codeaddict99: Okay, will recreate infura ID and rewrite down

packages/dapp/next.config.js
codeaddict99: Okay, will remove the env variables

sergfeldman:
This PR will not be reviewed.

PR #249, opened 2022-09-07
Refactor architecture using yarn workspace

cuddle-dev:


package.json
```diff
@@ -1,109 +1,35 @@
...
-    "_contracts:clean": "run-p _contracts:clean:*",
-    "_contracts:clean:dollar": "cd ./contracts/dollar && yarn clean",
-    "_contracts:clean:nft": "cd ./contracts/ubiquistick && yarn clean",
-    "contracts:run:nft": "cd ./contracts/ubiquistick && hardhat node",
-    "contracts:run:faucet": "cd ./contracts/dollar && yarn node:faucet",
-    "killscript": "kill `ps | grep dev-chain-init.sh | grep -v grep | awk '{ print $1 }'`",
-    "__format:cspell": "cspell '**'"
-  },
+  "name": "@ubiquity/monorepo",
+  "version": "1.0.1-beta",
```
0x4007, line 3: Would be dope to follow semver as part of our CI.

packages/dapp/components/lib/utils.ts
```diff
@@ -28,10 +27,10 @@ export async function performTransaction(transaction: Promise<ContractTransactio
   // return false;
 }
 
-export async function erc1155BalanceOf(addr: string, erc1155UbiquityCtr: ERC1155Ubiquity): Promise<BigNumber> {
+export async function erc1155BalanceOf(addr: string, erc1155UbiquityCtr: Contract): Promise<BigNumber> {
   const treasuryIds = await erc1155UbiquityCtr.holderTokens(addr);
 
-  const balanceOfs = treasuryIds.map((id) => {
+  const balanceOfs = treasuryIds.map((id: any) => {
```
0x4007: We should not be using `any` perhaps try using a more specific type like `number` or `string`. If none work then I would fall back to `unknown`.

I assumed that our commit hook would throw an error for this!

packages/dapp/components/redeem/UcrNftRedeem.tsx
```diff
@@ -43,9 +42,9 @@ const UcrNftRedeem = () => {
   if (!deployedContracts || !managedContracts || !debtIds) return <span>· · ·</span>;
   if (debtIds.length === 0) return <span>No uCR-NFT coupons</span>;
 
-  async function fetchDebts(address: string, contract: DebtCoupon) {
```
0x4007, line 46: The custom typing `DebtCoupon` has been replaced with a generic `Contract` typing. @0xcodercrane and I had a discussion on this in Telegram and agreed that these should be fixed/switched back in the future for enhanced type safety. I understand that there were complications to get these custom typings stable with Foundry.

Includes all the other similar changes to this.

0x4007 (approved):
Dope re-architecture good work @0xcodercrane. Telegram review transcript for future reference:

packages/dapp/components/redeem/UcrNftRedeem.tsx
cuddle-dev: created an issue for it.
https://github.com/ubiquity/ubiquity-dollar/issues/250

package.json
cuddle-dev: nice catch! but cspell doesn't work properly for some variable names. ie. curveuADBalance fails on spell check.
My suggestion is to make an issue if you actually want spell check with ignore words / paths and do it later

packages/dapp/components/lib/utils.ts
cuddle-dev: https://github.com/ubiquity/ubiquity-dollar/pull/249/commits/76d0b63e397ad3c7ef849d9f5c31f946ae6603c9

package.json
0x4007: > nice catch! but cspell doesn't work properly for some variable names. ie. curveuADBalance fails on spell check. My suggestion is to make an issue if you actually want spell check with ignore words / paths and do it later

#252

PR #255, opened 2022-09-12
Fix/lint staged

0x4007:
1. fix lint-staged (runs eslint and prettier on committed files only.)
2. move `/bin/` to `/utils/`
3. reimport nodemon config
4. reimport global dom typings (`*.svg` and `window.ethereum`)
5. run prettier and eslint on entire codebase (passes with ~60 warnings, no errors)
6. mobile sidebar (in the ui) enhancement. click off of sidebar to close it!


package.json
```diff
@@ -6,33 +6,43 @@
...
     "build:all": "yarn workspaces foreach run build",
     "test:all": "yarn workspaces foreach run test",
-    "purge:all": "yarn workspaces foreach run purge && rimraf './node_modules'",
+    "purge:all": "yarn workspaces foreach run purge && rimraf node_modules",
     "prettier:all": "yarn workspaces foreach run prettier",
     "lint:all": "yarn workspaces foreach run lint",
-    "build:ci:cloudflare": "yarn workspaces foreach run build",
-    "build:ci:netlify": "yarn workspaces foreach run build",
-    "build:ci:fleek": "yarn workspaces foreach run build"
+    "build:ci:cloudflare": "run-s build:all",
```
cuddle-dev: do we need to install run-s here? we can just use `yarn` like `yarn build:all`

cuddle-dev (approved):
Looks good to me

cuddle-dev:
can we temporarily stop changing for contracts workspaces as of now? because it has a bunch of refactors.

0x4007:
> can we temporarily stop changing for contracts workspaces as of now? because it has a bunch of refactors.

This is my first time using yarn workspaces but I presume that you're talking about changes in the `packages/contracts` directory.

cuddle-dev:
> > can we temporarily stop changing for contracts workspaces as of now? because it has a bunch of refactors.
> 
> This is my first time using yarn workspaces but I presume that you're talking about changes in the `packages/contracts` directory.

Right

package.json
0x4007: > do we need to install run-s here? we can just use `yarn` like `yarn build:all`

Okay good idea. So we can probably just do `    "build:ci:cloudflare": "build:all",` ?

cuddle-dev: yes. we can

PR #267, opened 2022-09-19
Feat/yield

### LINKED ISSUE #183: Proxy Yield Aggregator v1.0.0 - ERC4626 Compatible Deposit Contract
### Discussed in https://github.com/ubiquity/ubiquity-dollar-development/discussions/171

<div type='discussions-op-text'>

<sup>Originally posted by **pavlovcik** May 14, 2022</sup>
# "Bring Your Own Yield Source"

For the next iteration of the Proxy Yield Aggregator (PYA) program a user should be able to bring their own [ERC4626](https://github.com/fei-protocol/ERC4626/blob/7a947f2507b760ae470578cfb106f71ff5b1a14b/src/ERC4626Router.sol) compatible yield source, and connect it to our deposit contract. 

This product is intended for users to be able to stack the total yield that they can receive from their favorite yield bearing sources. 

Upon withdrawal, the user interface should present a slider which represents risk tolerance of the user. 

## Example

1. The user likes a new ERC4626 Compatible Yearn USDC vault that promises 20% APR.
1. The user "connects" the ERC4626 vault to our deposit contract. 
    1. Implementation details must be researched, but I hope that we can avoid a contract redeployment, a la deploying a new Uniswap market.
1. The user deposits 1000 USDC, and the USDC is immediately transferred to the underlying yield source.
1. One year later, the user seeks to withdraw their 1200 USDC, and returns to our user interface to withdraw.
1. The user interface presents a slider, which seeks to understand the user's risk/reward tolerance 
    1. For example, the user can withdraw with no additional risk (1200 USDC) or the user can withdraw with max risk (2400 uCR)
    1. The idea is that the user can convert the USDC into our uCR for a premium 
    1. In exchange, the Ubiquity Dollar is able to raise collateral for the dollar in perpetuity.
1. User withdraws at "25% r/r" netting them something like 900 USDC and 900 uCR (premium curve formula to be determined)
    1. "50% r/r" would be 600 USDC and 1200 uCR
    1. "100% r/r" is 0 USDC and 2400 uCR
    1. "16.7% r/r" is 1000 USDC and 400 uCR // I imagine that many would, in this example go here or lower in order to only convert yield, and to receive the principle back.

## Adjustable Parameters

1. uCR Premium, a math formula (default would be a linear calculation, 1 USDC -> 2 uCR)
    1. This would be deployed as a separate smart contract
    1. This gives us flexibility to implement exponential or logarithmic functions which may more efficiently produce our desired economic outcomes!
1. Yield token whitelist (default would be an array with the following contract addresses: USDC, DAI, LUSD)
    1. The idea is that, because the user is able to bring their own yield source, we wouldn't want a situation where they give assets that we do not want (like UST)
    1. Rejection should happen upon deposit, not withdrawal; in case we change our accepted assets after the user already has deposited their tokens.

## Remarks

It would be very interesting to match the uCR Premium (math formula) with the bond of the associated asset. That way we are purchasing at the same price premiums ecosystem wide. As long as we're mindful of 1. the interface implementation, and 2. that we architect it so we can associate the uCR Premium PER asset, then I think we can easily share the same formula between our bonds and the PYA. </div>

zgorizzo69:
added tests


cuddle-dev:
Great work! Looks good to me

cuddle-dev:
I think here are the last steps to get the remaining stuff done. @zgorizzo69 . hope this makes sense and its not much work for you. 🙏
- [ ] move packages/contracts/yield-aggregator/contracts into packages/contracts/src/yield-aggregator. If you just move them, there could be some conflicts in interfaces contracts. I hope you should be able to handle them
- [ ] move packages/contracts/yield-aggregator/test into packages/contracts/test/yield-aggregator. that would be great to fix interface/mock contract duplication issue
- [ ] Remove packages/contracts/dollar directory
- [ ] Remove packages/contracts/yield-aggregator directory

PR #274, opened 2022-09-29
fix #215

### LINKED ISSUE #215: Wrong Network Infinite UI Crash Loop
There's an infinite crash loop for when Metamask is connected on the wrong network (seems like anything but Ethereum mainnet)

The whole page seems to hot reload and it bogs down the browser.

image

21CenturyTarzan:
I fixed the network error issue. If a user choose wrong network, it automatically ask user to change network.


cuddle-dev:
I am not sure about why asking user to switch a network is sort of an issue. In general, it is needed to let users know which network our platform runs on. Another reason what I am thinking is that we just deployed smart contracts on ethereum network. so we don't want users to interact with other networks except mainnet. Would you @21CenturyTarzan give us more context/reason about your pull request like how does the `error` will generate user compliants?
@pavlovcik how do you think of this pr?

cuddle-dev:
oh well we didn't have that feature to ask users to switch network. I might be wrong above @21CenturyTarzan .

components/lib/hooks/useWeb3.tsx
```diff
@@ -57,28 +57,41 @@ export const UseWeb3Provider: React.FC<ChildrenShim> = ({ children }) => {
...
-          ...web3State,
-          connecting: false,
-          providerMode: "metamask",
-          provider: newProvider,
-          walletAddress: newWalletAddress,
-          signer: newSigner,
-        });
+
+      const chainId = await newProvider.getNetwork().then((network) => network.chainId);
+      if (chainId === 1) {
```
cuddle-dev: For hardcoded chainId, could we update it just like read it from .env? because the dapp should be able to work on any networks we deploy smart contracts on.

cuddle-dev (changes requested):
This is a nice catch!

components/lib/hooks/useWeb3.tsx
21CenturyTarzan: Yes, I can make it work from reading the .env and remove the hardcoded part.

21CenturyTarzan:
Hi, I removed the hardcoded part by using .env file.

components/lib/hooks/useWeb3.tsx
```diff
@@ -57,28 +58,41 @@ export const UseWeb3Provider: React.FC<ChildrenShim> = ({ children }) => {
...
-          walletAddress: newWalletAddress,
-          signer: newSigner,
-        });
+
+      const chainId = await newProvider.getNetwork().then((network) => network.chainId);
+      if (chainId === Chain_Id) {
+        setWeb3State({ ...web3State, connecting: true });
+        const addresses = (await newProvider.send("eth_requestAccounts", [])) as string[];
+        if (addresses.length > 0) {
+          console.log("Connected wallet ", addresses[0]);
```
cuddle-dev: lets just remove console.log and alert(Line# 92).

cuddle-dev (changes requested):
Looks good.

components/lib/hooks/useWeb3.tsx
21CenturyTarzan: Ok. No problem

21CenturyTarzan: Hi, I removed console.log and alert();

sergfeldman:
@21CenturyTarzan We will continue PR review next week.

cuddle-dev:
looks good to me @21CenturyTarzan . by the way, would you create a pull request for the development branch, not the main branch?

0x4007:
> looks good to me @21CenturyTarzan . by the way, would you create a pull request for the development branch, not the main branch?

fyi @0xcodercrane should be able to do this 
image

0x4007:
@21CenturyTarzan why did you make the mainnet chain id a variable? It should never change from `1`

21CenturyTarzan:
Ok. I will try to create pull request for the development branch.

cuddle-dev:
> @21CenturyTarzan why did you make the mainnet chain id a variable? It should never change from `1`

The dapp should be able to run on any networks except mainnet. so hardcoded values need to be removed.

0x4007:
> > @21CenturyTarzan why did you make the mainnet chain id a variable? It should never change from `1`
> 
> The dapp should be able to run on any networks except mainnet. so hardcoded values need to be removed.

Technically, I agree. But the only plans for official deployments currently are Mainnet (1) and Optimism (10)

cuddle-dev:
If you @21CenturyTarzan can resolve conflicts, that would be great and we will absolutely merge your pr

0x4007:
@21CenturyTarzan following up on this

Draeieg:
the wrong network button hasn't been added yet, currently it just results in a black screen

image

cuddle-dev:
we didn't get any response from @21CenturyTarzan. but we can make the simple bounty for this @pavlovcik @sergfeldman

PR #324, opened 2022-11-14
fix: show USDT balance in footer

### LINKED ISSUE #323: Tether Balance Not Rendering Correctly in Inventory
For once I have a Tether balance and noticed that its rendering as `0 USDT` in the Inventory even though I have exactly `1000 USDT`

image

image

ryzhak:
PR for https://github.com/ubiquity/ubiquity-dollar/issues/323

### DIFF
--- a/packages/dapp/components/layout/Inventory.tsx
+++ b/packages/dapp/components/layout/Inventory.tsx
@@ -50,7 +50,7 @@ const Inventory = () => {
 
   function showIfBalanceExists(key: keyof Balances | keyof typeof namedContracts, name: keyof typeof tokenSvg, id: string) {
     const usdcFix = function () {
-      if (key == "usdc") return 6;
+      if (key == "usdc" || key == "usdt") return 6;
       else return 18;
     };
 


packages/dapp/components/layout/Inventory.tsx
```diff
@@ -50,7 +50,7 @@ const Inventory = () => {
 
   function showIfBalanceExists(key: keyof Balances | keyof typeof namedContracts, name: keyof typeof tokenSvg, id: string) {
     const usdcFix = function () {
-      if (key == "usdc") return 6;
+      if (key == "usdc" || key == "usdt") return 6;
```
cuddle-dev, line 53: usdc/usdt decimals vary per chain for example the decimals of usdc is 18 on bnb chain. so it will not work on other chains.

cuddle-dev (changes requested):

packages/dapp/components/layout/Inventory.tsx
0x4007: I had no idea but is the correct solution to check the amount of decimals then render that amount of decimals every time?

I also think that this solution is fine for our purposes because we don't plan to support bsc. 

Basically only possibly roll ups that settle on Ethereum.

ryzhak:
so should I change smth in this PR?

0x4007:
I think its fine don't worry about it.

> so should I change smth in this PR?

PR #372, opened 2022-12-08
fix(apein): added multiple tokens deposit + improved tests

### LINKED ISSUE #370: improve DirectGovernanceFarmer to handle apeIn with four tokens
the deposit function only takes one token but we want people to be able to apeIn with up to four tokens instead of 
signing up to four transactions (deposit with uAD + deposit with USDC +deposit with DAI + deposit with USDT )

the goal is to add a deposit function that takes up to 4 tokens

zgorizzo69:
Resolves #370 

 I improved the tests to make sure  events is emitted

### DIFF
--- a/packages/contracts/src/dollar/DirectGovernanceFarmer.sol
+++ b/packages/contracts/src/dollar/DirectGovernanceFarmer.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity >=0.8.3;
+pragma solidity >=0.8.16;
 
 /**
  * @title Ubiquity.
@@ -37,7 +37,7 @@ contract DirectGovernanceFarmer is ReentrancyGuard {
     );
     event DepositMultiple(
         address indexed sender,
-        uint256[4] amount,
+        uint256[4] amounts,
         uint256 durationWeeks,
         uint256 stakingShareId
     );
@@ -47,6 +47,11 @@ contract DirectGovernanceFarmer is ReentrancyGuard {
         uint256 stakingShareId,
         address token,
         uint256 amount
+    );
+    event WithdrawAll(
+        address indexed sender,
+        uint256 stakingShareId,
+        uint256[4] amounts
     );
 
     constructor(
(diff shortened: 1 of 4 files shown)


packages/contracts/src/dollar/DirectGovernanceFarmer.sol
```diff
@@ -43,8 +49,11 @@ contract DirectGovernanceFarmer is ReentrancyGuard {
         uint256 amount
     );
 
-
-    constructor(address _manager, address base3Pool, address depositZap) {
+    constructor(
```
hashedMae, line 57: `onERC1155Received` should check that the operator is authorized, from matches user, and ID exists

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
```diff
@@ -90,42 +98,158 @@ contract DirectGovernanceFarmer is ReentrancyGuard {
...
+
+    /**
+     * @dev Deposit into Ubiquity protocol
+     * @notice Stable coins (DAI / USDC / USDT / Ubiquity Dollar) => uAD3CRV-f => Ubiquity StakingShare
+     * @notice STEP 1 : Change (DAI / USDC / USDT / Ubiquity dollar) to 3CRV at uAD3CRV MetaPool
+     * @notice STEP 2 : uAD3CRV-f => Ubiquity StakingShare
+     * @param tokenAmounts Amount of tokens to deposit (For max: `uint256(-1)`) it MUST follow this order [Ubiquity Dollar, DAI, USDC, USDT]
+     * @param durationWeeks Duration in weeks tokens will be locked (1-208)
+     */
+    function deposit(uint256[4] calldata tokenAmounts, uint256 durationWeeks)
```
hashedMae: There's two functions called Deposit. They'll have different signatures but should still have different names.

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
```diff
@@ -136,60 +260,96 @@ contract DirectGovernanceFarmer is ReentrancyGuard {
      * @param stakingShareId Staking Share Id to withdraw
      * @param token Token to withdraw to : DAI, USDC, USDT, 3CRV or Ubiquity Dollar
      */
-    function withdraw(
-        uint256 stakingShareId,
-        address token
-    ) external nonReentrant returns (uint256 tokenAmount) {
-        
+    function withdraw(uint256 stakingShareId, address token)
```
hashedMae: Limiting user to withdraw only the coin is quite limiting and forces them to incur some slippage

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity >= 0.8.3;
+pragma solidity >=0.8.3;
```
hashedMae: I believe we're using solidity 0.8.16 now.

zgorizzo69: done

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
zgorizzo69: we can have different signatures with the same name. why do you suggest to have another name ? what name do you suggest ?

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
zgorizzo69: not sure what it is for,  I am not the one who created that contract and the goal here is to add another deposit function maybe you should create another issue if we have to accept erc1155 tokens

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
zgorizzo69: also not really part of that issue but it makes sense to let people withdraw all tokens (why wasn't part of the first version btw ?). I will add it nonetheless

0x4007: The history of this contract is that 1. zapaz wrote most of the original logic for use with InstaDapp, and then 2. we had a bounty hunter take that contract, remove the InstaDapp specific code, and create this current version.

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
hashedMae: It makes code more readable and can clarify the difference between the two deposit functions without having to read into how they perform differently. 
I would use depositSingle and depositMulti

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
hashedMae: K, out of spec for this issue then.

packages/contracts/src/dollar/DirectGovernanceFarmer.sol
zgorizzo69: I will update it but can be an interesting discussion for Christmas eve " Pros and Cons of function overloading"

0x4007: > I will update it but can be an interesting discussion for Christmas eve " Pros and Cons of function overloading"

Classic Christmas dinner discussion topics.

Issue #377, opened 2022-12-09
Build fails on a fresh yarn.lock


rndquu:
On a fresh install there is a build error `Module not found: Can't resolve 'jotai-immer'`

How to reproduce:
1. Delete `yarn.lock`
2. Run `yarn install`
3. Run `yarn build:all`

Gives the following:
```
➤ YN0000: No files changed, compilation skipped
➤ YN0000: info  - Linting and checking validity of types...
➤ YN0000: warn  - The Next.js plugin was not detected in your ESLint configuration. See https://nextjs.org/docs/basic-features/eslint#migrating-existing-config
➤ YN0000: 
➤ YN0000: ./pages/404.tsx
➤ YN0000: 12:82  Warning: Replace `Docs` with `⏎··············Docs⏎············`  prettier/prettier
➤ YN0000: 15:56  Warning: Replace `FAQ` with `⏎··············FAQ⏎············`  prettier/prettier
➤ YN0000: 18:77  Warning: Replace `Github` with `⏎··············Github⏎············`  prettier/prettier
➤ YN0000: 21:63  Warning: Replace `Discord` with `⏎··············Discord⏎············`  prettier/prettier
➤ YN0000: 24:58  Warning: Replace `Telegram` with `⏎··············Telegram⏎············`  prettier/prettier
➤ YN0000: 27:65  Warning: Replace `Twitter` with `⏎··············Twitter⏎············`  prettier/prettier
➤ YN0000: 
➤ YN0000: ./pages/index.tsx
➤ YN0000: 13:10  Warning: 'twapPrice' is assigned a value but never used.  @typescript-eslint/no-unused-vars
➤ YN0000: 
➤ YN0000: ./components/bonds/App.tsx
➤ YN0000: 36:10  Warning: 'showAdminComponents' is assigned a value but never used.  @typescript-eslint/no-unused-vars
➤ YN0000: 
➤ YN0000: ./components/lib/hooks/useDeployedAddress.tsx
➤ YN0000: 6:41  Warning: Unexpected any. Specify a different type.  @typescript-eslint/no-explicit-any
➤ YN0000: 
➤ YN0000: ./components/lib/hooks/useWeb3.tsx
➤ YN0000: 9:7  Warning: 'LOCAL_NODE_ADDRESS' is assigned a value but never used.  @typescript-eslint/no-unused-vars
➤ YN0000: 
➤ YN0000: ./components/redeem/UcrRedeem.tsx
[...]

PR #353, opened 2022-11-30
build: add conventional commits

rndquu:
PR for https://github.com/ubiquity/ubiquity-dollar/issues/344

This PR enables check for commit messages.

Before this PR linter was not running in pre-commit hook, but now pre-commit hook is enabled and linter is running. I commented linter code because:
1. it is not clear what is the best strategy to run the linter
2. reenabling linter is out of scope of this task

PR #378, opened 2022-12-09
build: add conventional commits

### LINKED ISSUE #344: Enforce Conventional Commits
The following is commented away in our README but we never implemented this feature. It is a great way to standardize commit messages as well as to produce a great standardized input for automation tools, which for example, can automatically produce change logs for every deployment.

> 1. We [automatically enforce](https://github.com/conventional-changelog/commitlint) the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format for commit messages. This can be frustrating for junior developers, but I promise that you'll quickly get used to it! -->

> The Conventional Commits specification is a lightweight convention on top of commit messages. It provides an easy set of rules for creating an explicit commit history; which makes it easier to write automated tools on top of. -->

rndquu:
Resolves #344

This PR enables check for commit messages.

There can be 2 flows:
1. When a new user git clones the project and runs `yarn install` then commit msg hook is installed in this `postinstall` script https://github.com/ubiquity/ubiquity-dollar/compare/development...rndquu:issue/344?expand=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R35
2. Existing users (who have already git cloned the project) have to manually run `yarn postinstall` in order to install commit msg hooks

Before this PR linter was not running in pre-commit hook, but now pre-commit hook is enabled and linter is running. I commented linter code because:
1. it is not clear what is the best strategy to run the linter
2. reenabling linter is out of scope of this task

P.S. I closed the previous PR https://github.com/ubiquity/ubiquity-dollar/pull/353 because it has conflicts in `yarn.lock` and after solving the conflicts the build fails due to https://github.com/ubiquity/ubiquity-dollar/issues/377

### DIFF
--- a/.github/workflows/enforce-conventional-commits.yml
+++ b/.github/workflows/enforce-conventional-commits.yml
@@ -1,0 +1,14 @@
+name: Conventional Commits
+
+on:
+  pull_request:
+    branches: [ development ]
+
+jobs:
+  build:
+    name: Conventional Commits
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: webiny/action-conventional-commits@v1.1.0
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,3 +1,5 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-yarn lint-staged --verbose
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+# NOTICE: temporary disabled because it is not clear what is the best way to lint
+# yarn lint-staged --verbose
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-# NOTICE: temporary disabled because it is not clear what is the best way to lint
-# yarn lint-staged --verbose
+yarn lint-staged --verbose
--- a/package.json
+++ b/package.json
@@ -9,12 +9,15 @@
     "packages/contracts"
   ],
   "devDependencies": {
+    "@commitlint/cli": "^17.3.0",
+    "@commitlint/config-conventional": "^17.3.0",
     "@typescript-eslint/eslint-plugin": "^5.32.0",
     "@typescript-eslint/parser": "^5.32.0",
     "eslint": "^8.23.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-tsc": "^2.0.0",
+    "husky": "^8.0.2",
     "lint-staged": "^13.0.3",
     "nodemon": "^2.0.19",
     "rimraf": "3.0.2"
@@ -28,13 +31,8 @@
     "prettier:all": "yarn workspaces foreach run prettier",
     "lint:all": "yarn workspaces foreach run lint",
     "build:ci:cloudflare": "bash ./utils/build_cloudflare.sh",
-    "build:ci:netlify": "yarn build:all"
-  },
-  "husky": {
-    "hooks": {
-      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "lint-staged"
-    }
+    "build:ci:netlify": "yarn build:all",
+    "postinstall": "husky install"
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx,json,sol}": [
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+yarn commitlint --edit $1
--- a/packages/dapp/package.json
+++ b/packages/dapp/package.json
@@ -13,8 +13,6 @@
     "prettier": "prettier --write ."
   },
   "devDependencies": {
-    "@commitlint/cli": "^17.1.2",
-    "@commitlint/config-conventional": "^17.1.0",
     "@ethersproject/abi": "^5.0.0",
     "@ethersproject/bytes": "^5.0.0",
     "@ethersproject/providers": "^5.6.8",
@@ -41,7 +39,6 @@
     "connectkit": "^1.1.0",
     "ethers": "^5.7.2",
     "hardhat": "^2.10.1",
-    "husky": "^8.0.1",
     "next": "^12.2.3",
     "npm-run-all": "^4.1.5",
     "react": "^18.2.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -278,32 +278,32 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@commitlint/cli@npm:^17.1.2":
-  version: 17.1.2
-  resolution: "@commitlint/cli@npm:17.1.2"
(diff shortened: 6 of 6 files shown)


0x4007:
I wish there was a better solution for 2. Any ideas on how to enforce this? Perhaps we can on CI? Just make sure that the git hook is automatically installed after the developer runs `yarn` to install dependencies. That's quite common to do after a pull.

zgorizzo69:
@rndquu anyway people can still bypass it if they want so for me it is fine we just have to check that commits are ok from time to time

0x4007:
> @rndquu anyway people can still bypass it if they want so for me it is fine we just have to check that commits are ok from time to time

I believe that we could fork a script somewhere (instead of handwriting a new one) to enforce in CI

.husky/pre-commit
```diff
@@ -1,3 +1,5 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-yarn lint-staged --verbose
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+# NOTICE: temporary disabled because it is not clear what is the best way to lint
+# yarn lint-staged --verbose
```
0x4007: I don't understand what you mean by "best way to lint" 

It runs on changed files using `lint-staged`

It should load the settings we already defined.

package.json
```diff
@@ -28,13 +31,8 @@
     "prettier:all": "yarn workspaces foreach run prettier",
     "lint:all": "yarn workspaces foreach run lint",
     "build:ci:cloudflare": "bash ./utils/build_cloudflare.sh",
-    "build:ci:netlify": "yarn build:all"
-  },
-  "husky": {
-    "hooks": {
-      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "lint-staged"
```
0x4007, line 36: I assume that this is needed for `lint-staged` to run?

.husky/pre-commit
rndquu: `lint-staged` is enabled again

I meant 2 things:
1. enabling linter is out of scope for this task
2. I personally don't like when code is updated via automatic tools like `prettier --write` or especially `eslint --fix`. By "best way to lint" I meant relaxing linting rules as much as possible.

package.json
rndquu: No, script command is run from https://github.com/ubiquity/ubiquity-dollar/blob/8f295b7eb0983254b26b5f3dab9f480a88831b96/.husky/pre-commit

I tried to remove husky config from `package.json` and everything worked fined so I decided not to keep commands in 2 places: `package.json` and `.husky` folder thus leaving only `.husky` folder.

rndquu:
@pavlovcik @zgorizzo69 

udpates:
1. linter is enabled
2. added github action for checking commit msgs

So now if user skips msg check on commit via `--no-verify` then CI will block the PR

.github/workflows/enforce-conventional-commits.yml
```diff
@@ -0,0 +1,14 @@
+name: Conventional Commits
+
+on:
+  pull_request:
+    branches: [ development ]
```
0x4007, line 5:
```suggestion
    branches: * 
```

I'm not sure if this is the correct syntax but we should enforce on every branch because what if a bounty hunter has a main branch for their PR and then sub branches with the wrong commit styles. Wouldn't those commits be difficult to parse for any automated tooling down the line?

0x4007 (changes requested):

0x4007:
I guess its hard for me to verify if the CI works because CI only runs from `development` 

I suppose we could always merge and then finish on `development` if needed.

What happens if the bounty hunter has a ton of non-compliant commits previously? Does CI only check the latest one? If they have to rename all of their previous commits that could be terrible but I can't think of a better solution.

.github/workflows/enforce-conventional-commits.yml
rndquu: Currently the merge flow is the following:
1. Bounty hunters create PRs to the `development` branch
2. You manually merge `development` into `main` from time to time

There is no way to miss invalid commit msg because all PRs are getting in the `development` branch first

rndquu:
> I guess its hard for me to verify if the CI works because CI only runs from development

CI action runs on PR to the `development` branch. So you pushed a commit https://github.com/ubiquity/ubiquity-dollar/pull/378/commits/6f0e4d2116d651af6122bc48c18feb01b35e5733 which expectedly failed the CI action because of invalid commit msg name

> Does CI only check the latest one? 

it runs on all commit messages

> What happens if the bounty hunter has a ton of non-compliant commits previously?

This will be hard to solve. Bounty hunter will have to squash all PR commits into a single one with a valid commit msg, smth like `git reset --soft HEAD~3 && git commit -m "feat: new feature"`

0x4007:
> This will be hard to solve. Bounty hunter will have to squash all PR commits into a single one with a valid commit msg, smth like `git reset --soft HEAD~3 && git commit -m "feat: new feature"`

I have a feeling that this will be an issue in the future, but again I can't figure out a better solution for this 😰

0x4007:
Can you demonstrate what you need to do here?

I hope that there is a way to fix this issue without deleting previous commits! PR reviews are much easier without deletes/force-pushes

I enabled auto-merge so assuming that your next commit passes, you will be able to merge this in to complete the bounty!

.github/workflows/enforce-conventional-commits.yml
pavlovcik: I would much rather have the bounty hunters handle the squash merge / fix commit message issue than me!

package.json
0x4007: Awesome work!

0x4007 (approved):

rndquu:
@pavlovcik 

FYI, this PR is somehow automerged even with failed CI check

1. There were 2 CI issues: unresolved conversations (https://github.com/ubiquity/ubiquity-dollar/pull/378#discussion_r1045203487) and failed CI action "Conventional Commits" because of your test commit with invalid message (https://github.com/ubiquity/ubiquity-dollar/pull/378/commits/6f0e4d2116d651af6122bc48c18feb01b35e5733)
2. Then I clicked "resolve conversation" and PR automatically merged even with failed CI action

This is a pretty weird behaviour of github or I just don't understand smth about "auto merges" 

Anyway the PR was ready for merging

.husky/pre-commit
0x4007: Looks like you disabled it again.

0x4007:
Looks like this became a bit of a mess. 

You disabled the prettier again on the latest commit in the pre-commit hook, which should remain enabled. It's still unclear to me why you want to remove it. I think that linter/prettier should not be loosely enforced because then we will accrue formatting issues over time. Since the tools are automated, it shouldn't be burdensome for the developers working on the code either.

It looks like the CI didn't run on your latest commit so I presume that it only checks on the merge PR event instead of the commit event (like when you write a GitHub Action and how it is supposed to run "on pull_request" or "on commit") which could be after your CI is supposed to run or something. I am unsure.

rndquu:
> You disabled the prettier again on the latest commit in the pre-commit hook, which should remain enabled

Where do you see it? This line https://github.com/ubiquity/ubiquity-dollar/blob/development/.husky/pre-commit#L4 is uncommented + I've just checked the latest build from `development` branch and `lint-staged` is running as expected on commit.

0x4007:
I realized I was looking at your first commit, not your latest - sorry!

PR #313, opened 2022-11-01
ERC-2535 Diamond Implementation

0xautohub:
- [x] Full research on Diamond
- [x] Smart contract Implementation while paying attention to access control system
- [x] Unit Tests
- [x] Deployment Script
- [x] Full Test on Goerli test network and the [Ethereum Diamond Inspector](https://louper.dev/)

PR #380, opened 2022-12-10
Added diamond implementation, unit test as well as deployment script

### LINKED ISSUE #300: Upgrade `UbiquityAlgorithmicDollarManager.sol` to ERC2535 Diamond Standard
We have limited slots to associate contracts with our protocol on the current Manager implementation. The Diamond Standard seems to be a useful architecture for our protocol as we continue to plan to build an ecosystem around our dollar.

This project would require the smart contract implementation, associated tests, and finally a redeployment. 

As of this moment it is not clear if we must redeploy everything, or if we can simply redeploy the Manager and use the new Manager instead.

0xautohub:
Resolves #300

main PR was [**PR 313**](https://github.com/ubiquity/ubiquity-dollar/pull/313)

Applied current product names on the **PR 313** 

https://goerli.etherscan.io/address/0x69AB9Aa8a8C38c276E389d6EfaDE710B16D91100
https://louper.dev/diamond/0x69AB9Aa8a8C38c276E389d6EfaDE710B16D91100?network=goerli


0xautohub:
@pavlovcik @0xcodercrane This PR is ready for a review

packages/contracts/package.json
```diff
@@ -28,6 +28,7 @@
     "echidna": "echidna-test . --config echidna.config.yml",
     "prettier": "forge fmt",
     "deploy": "ts-node ./scripts/deploy/deploy.ts",
+    "deploy:diamond": "ts-node ./scripts/deploy/manager/DiamondInAll.ts",
```
0x4007, line 31: Intuitively I feel that this doesn't belong here but I don't have any better ideas.

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
```diff
@@ -0,0 +1,195 @@
...
+        }
+        flattenConstructorArgs += `]\"`;
+        prepareCmd = `forge create --json --rpc-url ${args.rpcUrl} --private-key ${args.privateKey} ${args.contractInstance} --constructor-args ${flattenConstructorArgs}`;
+    } else {
+        prepareCmd = `forge create --json --rpc-url ${args.rpcUrl} --private-key ${args.privateKey} ${args.contractInstance}`;
+    }
+
+    const chainId = Networks[args.network] ?? undefined;
+    if (!chainId) {
+        throw new Error(`Unsupported network: ${args.network} Please configure it out first`);
```
0x4007:
```suggestion
        throw new Error(`Unsupported network: ${args.network}.`);
```

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
```diff
@@ -0,0 +1,195 @@
...
+        stdout = err?.stdout;
+        stderr = err?.stderr;
+    }
+    if (stdout) {
+        const regex = /{(?:[^{}]*|(R))*}/g;
+        const found = stdout.match(regex);
+        if (found && JSON.parse(found[0])?.deployedTo) {
+            const { deployedTo, deployer, transactionHash } = JSON.parse(found[0]);
+            result = { deployedTo, deployer, transactionHash };
+            console.log(`Deployed ${args.name} contract successfully. res: ${deployedTo}`);
```
0x4007:
```suggestion
            console.log(`Deployed ${args.name} contract successfully. Address: ${deployedTo}`);
```

packages/contracts/scripts/deploy/manager/PrintSelectors.ts
```diff
@@ -0,0 +1,44 @@
...
+  const contractArtifact = require(contractFilePath);
+  const abi = contractArtifact.abi;
+  const bytecode = contractArtifact.bytecode;
+  const target = new ethers.ContractFactory(abi, bytecode);
+  const signatures = Object.keys(target.interface.functions);
+
+  const selectors = signatures.reduce((acc, val) => {
+    if (val !== "init(bytes)") {
+      // @ts-ignore
+      acc.push(target.interface.getSighash(val));
```
0x4007, line 28: Is this going to cause issues? I don't know Foundry or Solidity well.

packages/contracts/src/dollar/mocks/MockGovernanceToken.sol
```diff
@@ -0,0 +1,19 @@
+// contracts/GLDToken.sol
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.3;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract MockGovernanceToken is ERC20 {
```
0x4007: What is the purpose of a mock governance token in the same repo with the real governance token? Why not just use the real one?

0x4007:
It all looks pretty good to me. Let's see what the Solidity experts think!

packages/contracts/scripts/deploy/manager/PrintSelectors.ts
0xautohub: > Is this going to cause issues? I don't know Foundry or Solidity well.

In diamond, function with the same name can exist only one, even they are on difference facets
And it's important to check if there is any duplicated function before adding a new facet or new functions
PrintSelectors.ts file is used to print the function selectors of any facet

packages/contracts/src/dollar/mocks/MockGovernanceToken.sol
0xautohub: > What is the purpose of a mock governance token in the same repo with the real governance token? Why not just use the real one?

Originally, I thought I need ERC20 contract with transfer() function, but just found that it can be removed

sergfeldman:
@0xcodercrane @zgorizzo69 @hashedMae Please do a thorough review of this important update

sergfeldman:
Ubiquity is implementing the EIP-2535 Diamond.
We offer incentives for a thorough PR review.
@mudgen @Teamo0 @rsundar Would you be interested in performing the PR review?

packages/contracts/src/manager/interfaces/IDiamondCut.sol
```diff
@@ -0,0 +1,31 @@
...
+    /// @param _init The address of the contract or facet to execute _calldata
+    /// @param _calldata A function call, including function selector and arguments
+    ///                  _calldata is executed with delegatecall on _init
+    function diamondCut(
+        FacetCut[] calldata _diamondCut,
+        address _init,
+        bytes calldata _calldata
+    ) external;
+
+    event DiamondCut(FacetCut[] _diamondCut, address _init, bytes _calldata);
```
zgorizzo69, line 30: sometimes events are declared at the beginning sometimes at the end
IMO solhint would enforce a unified coding style + warn about some problems

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
```diff
@@ -0,0 +1,195 @@
...
+            const { deployedTo, deployer, transactionHash } = JSON.parse(found[0]);
+            result = { deployedTo, deployer, transactionHash };
+            console.log(`Deployed ${args.name} contract successfully. Address: ${deployedTo}`);
+        }
+    }
+    return { result, stderr }
+}
+
+async function deployDiamond() {
+    const envPath = path.join(__dirname, "../../../.env");
```
zgorizzo69, line 107: nice job ! :+1:

packages/contracts/src/manager/upgradeInitializers/DiamondInit.sol
```diff
@@ -0,0 +1,44 @@
...
+
+        _grantRole(AccessControlStorage.DEFAULT_ADMIN_ROLE, _admin);
+        _grantRole(GOVERNANCE_TOKEN_MINTER_ROLE, _admin);
+        _grantRole(PAUSER_ROLE, _admin);
+        _grantRole(CREDIT_NFT_MANAGER_ROLE, _admin);
+        _grantRole(STAKING_MANAGER_ROLE, _admin);
+        _grantRole(INCENTIVE_MANAGER_ROLE, _admin);
+        _grantRole(GOVERNANCE_TOKEN_MANAGER_ROLE, address(this));
+
+        s.dollarTokenAddress = address(this);
```
zgorizzo69: if the dollar is part of the diamond we should add :
- ds.supportedInterfaces[type(IERC20Ubiquity).interfaceId] = true;
- the dollar facet in the deployment script

packages/contracts/src/manager/facets/ManagerFacet.sol
```diff
@@ -0,0 +1,251 @@
...
+    {
+        s.twapOracleAddress = _twapOracleAddress;
+        // to be removed
+
+        TWAPOracleDollar3pool oracle =
+            TWAPOracleDollar3pool(s.twapOracleAddress);
+        oracle.update();
+    }
+
+    function setDollarTokenAddress(address _dollarTokenAddress)
```
zgorizzo69, line 29: In the diamond Init it is pat of the diamond so we don't need it here as it will always be the diamond address

packages/contracts/src/manager/facets/ManagerFacet.sol
```diff
@@ -0,0 +1,251 @@
...
+contract ManagerFacet is Modifiers {
+    // TODO Add a generic setter for extra addresses that needs to be linked
+    function setTwapOracleAddress(address _twapOracleAddress)
+        external
+        onlyAdmin
+    {
+        s.twapOracleAddress = _twapOracleAddress;
+        // to be removed
+
+        TWAPOracleDollar3pool oracle =
```
zgorizzo69, line 24: why the oracle isn't part of the diamond ?

packages/contracts/src/manager/facets/ManagerFacet.sol
```diff
@@ -0,0 +1,251 @@
...
+        external
+        onlyAdmin
+    {
+        s.dollarTokenAddress = _dollarTokenAddress;
+    }
+
+    function setCreditTokenAddress(address _creditTokenAddress)
+        external
+        onlyAdmin
+    {
```
zgorizzo69, line 39: all tokens can be part of the diamond and thus we can remove them from this storage

packages/contracts/src/manager/facets/ManagerFacet.sol
```diff
@@ -0,0 +1,251 @@
...
+        s.creditNFTCalculatorAddress = _creditNFTCalculatorAddress;
+    }
+
+    function setDollarMintCalculatorAddress(
+        address _dollarMintCalculatorAddress
+    ) external onlyAdmin {
+        s.dollarMintCalculatorAddress = _dollarMintCalculatorAddress;
```
zgorizzo69, line 81: calculators are also a good fit for a facet

packages/contracts/src/manager/libraries/LibAppStorage.sol
```diff
@@ -0,0 +1,62 @@
...
+    keccak256("GOVERNANCE_TOKEN_BURNER_ROLE");
+bytes32 constant PAUSER_ROLE = keccak256("PAUSER_ROLE");
+bytes32 constant CREDIT_NFT_MANAGER_ROLE = keccak256("CREDIT_NFT_MANAGER_ROLE");
+bytes32 constant STAKING_MANAGER_ROLE = keccak256("STAKING_MANAGER_ROLE");
+bytes32 constant INCENTIVE_MANAGER_ROLE = keccak256("INCENTIVE_MANAGER");
+bytes32 constant GOVERNANCE_TOKEN_MANAGER_ROLE =
+    keccak256("GOVERNANCE_TOKEN_MANAGER_ROLE");
+
+struct AppStorage {
+    address twapOracleAddress;
```
zgorizzo69, line 20: I would add a warning inside the comment here like
/* DO NOT CHANGE THE ORDER ONLY APPEND AT THE END*/

packages/contracts/src/manager/libraries/LibAppStorage.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
zgorizzo69, line 2: elsewhere it is 0.8.8 can you push all of them to 0.8.16 ?

packages/contracts/test/helpers/DiamondTestHelper.sol
```diff
@@ -0,0 +1,174 @@
...
+import "forge-std/Test.sol";
+import "forge-std/Vm.sol";
+import "forge-std/console.sol";
+import "../../src/manager/interfaces/IDiamondCut.sol";
+import "../../src/manager/interfaces/IDiamondLoupe.sol";
+
+contract DiamondTestHelper is IDiamondCut, IDiamondLoupe, Test {
+    uint256 private seed;
+
+    modifier prankAs(address caller) {
```
zgorizzo69, line 13: nice :+1:

packages/contracts/test/manager/facets/ManagerFacet.t.sol
```diff
@@ -0,0 +1,203 @@
...
+            address(dollarToken),
+            address(0x101),
+            100,
+            100
+        );
+        IManagerFacet.setTwapOracleAddress(address(_twapOracle));
+        assertEq(IManagerFacet.getTwapOracleAddress(), address(_twapOracle));
+    }
+
+    function testShouldSetDollarTokenAddress() public prankAs(admin) {
```
zgorizzo69, line 32: can we have a test that assert we can't set if not admin ?

packages/contracts/src/manager/libraries/LibAppStorage.sol
```diff
@@ -0,0 +1,62 @@
...
+}
+
+contract Modifiers is AccessControl {
+    AppStorage internal s;
+
+    modifier onlyOwner() {
+        LibDiamond.enforceIsContractOwner();
+        _;
+    }
+
```
zgorizzo69, line 54: I would add a non reentrant modifier here cos we will need it sooner or later and we can have these fields added to the appstorage struct 
```
   // reentrancy guard
    uint256 NOT_ENTERED;
    uint256 ENTERED;
    uint256 reentrancyStatus;
```

packages/contracts/test/manager/facets/ManagerFacet.t.sol
```diff
@@ -0,0 +1,203 @@
...
+
+import "../DiamondTestSetup.sol";
+import "../../../src/dollar/interfaces/ICurveFactory.sol";
+import "../../../src/dollar/interfaces/IMetaPool.sol";
+import "../../../src/dollar/mocks/MockDollarToken.sol";
+import "../../../src/dollar/mocks/MockTWAPOracleDollar3pool.sol";
+
+import {GOVERNANCE_TOKEN_MINTER_ROLE, GOVERNANCE_TOKEN_BURNER_ROLE} from "../../../src/manager/libraries/LibAppStorage.sol";
+
+contract TestManagerFacet is DiamondSetup {
```
zgorizzo69, line 12: IMO we need other tests 
- one where we add a facet that reads from appstorage
- one that add a field to appstorage struct 
- one we a new facet that need an init function call

zgorizzo69 (changes requested):

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
```diff
@@ -0,0 +1,195 @@
...
+        const regex = /{(?:[^{}]*|(R))*}/g;
+        const found = stdout.match(regex);
+        if (found && JSON.parse(found[0])?.deployedTo) {
+            const { deployedTo, deployer, transactionHash } = JSON.parse(found[0]);
+            result = { deployedTo, deployer, transactionHash };
+            console.log(`Deployed ${args.name} contract successfully. Address: ${deployedTo}`);
+        }
+    }
+    return { result, stderr }
+}
```
cuddle-dev, line 104: Why couldn't you use `packages/contracts/scripts/deploy/create.ts` instead of making a new one which is almost same?  Also facet deployments information should be recorded As of now thats fine to have them recorded in deployments.json. In the future, we will also compare `bytecode` to avoid redeployments of same codebase. but it will definitely be the next task. Lets use the origin `create` function. Its also allowed to update it if needed.

packages/contracts/package.json
cuddle-dev, line 31: https://goerli.etherscan.io/tx/0x3a21916c1b2485bca894ee28dc6f00591cd070e6899fd95f0ebd34d701e37f8f
https://goerli.etherscan.io/tx/0x184a06d79dd58ea5fea7082e27763942af6aa9353060ccfe8e23ac039194e7e9

yarn workspace @ubiquity/contracts run deploy:diamond is failing at the last step. Please look at above 2 txs I've tried on your branch.

Reply to : https://github.com/ubiquity/ubiquity-dollar/pull/313#issuecomment-1341259363
> diamondCut function would handle all upgrading of the contract, no need to add upgrade script, imo

In terms of smart contract side, DiamondCutFacet::diamondCut will handle add/replace/remove actions properly but what we want here is high-level script. `deploy:diamond` will redeploy everything each time we  try to deploy which wouldn't be useful in real environment.

cuddle-dev (changes requested):
https://github.com/ubiquity/ubiquity-dollar/pull/313#pullrequestreview-1207634171

I want to re-mention my last requests. (Let me ignore 3rd request bc it's less important)

**upgrade script, upgrade txhash, replace all the usages of original UADManager**   
3 checkpoints still not done.

hashedMae:
Where were the Diamond Libraries sourced from? Those should also be verified and if they've been audited we should review those as well.

packages/contracts/src/manager/access/AccessControl.sol
```diff
@@ -0,0 +1,58 @@
...
+
+import {IAccessControl} from "../interfaces/IAccessControl.sol";
+import {AccessControlInternal} from "./AccessControlInternal.sol";
+
+/**
+ * @title Role-based access control system
+ * @dev derived from https://github.com/OpenZeppelin/openzeppelin-contracts (MIT license)
+ * @dev https://github.com/solidstate-network/solidstate-solidity/blob/master/contracts/access/access_control/AccessControl.sol
+ */
+abstract contract AccessControl is IAccessControl, AccessControlInternal {
```
hashedMae, line 13: Unless we forking and maintaining our own version of these contracts then they should just be imported from lib/open-zeppelin. There's no reason to copy them here.

packages/contracts/src/manager/libraries/LibDiamond.sol
```diff
@@ -0,0 +1,211 @@
...
+import { IDiamondLoupe } from "../interfaces/IDiamondLoupe.sol";
+import "@openzeppelin/contracts/interfaces/IERC165.sol";
+import { IERC173 } from "../interfaces/IERC173.sol";
+
+// Remember to add the loupe functions from DiamondLoupeFacet to the diamond.
+// The loupe functions are required by the EIP2535 Diamonds standard
+
+error InitializationFunctionReverted(address _initializationContractAddress, bytes _calldata);
+
+library LibDiamond {
```
hashedMae, line 14: All dependencies should be added through `forge install` and not copy paste contracts unless we are forking them and plan to maintain them ourselves.

packages/contracts/src/manager/interfaces/IDiamondCut.sol
hashedMae: [Solidity Style Guide](https://docs.soliditylang.org/en/v0.8.17/style-guide.html#order-of-layout) is the best reference in this case

packages/contracts/src/manager/interfaces/IDiamondCut.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IDiamondCut {
```
hashedMae, line 4: According to the spec this should be implemented within IDiamond and not IDiamondCut

packages/contracts/src/manager/Diamond.sol
```diff
@@ -0,0 +1,49 @@
...
+import {DiamondLoupeFacet} from "./facets/DiamondLoupeFacet.sol";
+import {OwnershipFacet} from "./facets/OwnershipFacet.sol";
+
+struct DiamondArgs {
+    address owner;
+    address init;
+    bytes initCalldata;
+}
+
+contract Diamond {
```
hashedMae, line 19: Should inherit IDiamond

packages/contracts/src/manager/Diamond.sol
```diff
@@ -0,0 +1,49 @@
...
+    constructor(
+        DiamondArgs memory _args,
+        IDiamondCut.FacetCut[] memory _diamondCutFacets
+    ) {
+        LibDiamond.setContractOwner(_args.owner);
+        LibDiamond.diamondCut(_diamondCutFacets, _args.init, _args.initCalldata);
```
hashedMae, line 25: Should emit the `DiamondCut` event

packages/contracts/test/manager/facets/ManagerFacet.t.sol
hashedMae: That would be testing the AccessControl contracts which is outside of the scope. If we're using untrusted libraries then those should be audited and tested before being implemented anywhere.

packages/contracts/test/manager/facets/ManagerFacet.t.sol
```diff
@@ -0,0 +1,203 @@
...
+    function testCanCallGeneralFunctions() public {
+        IManagerFacet.getExcessDollarsDistributor(contract1);
```
hashedMae, line 14: This test isn't testing any assertions, it's only calling a function and showing that it won't revert.

packages/contracts/test/manager/DiamondTest.t.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./DiamondTestSetup.sol";
+
+contract TestDiamond is DiamondSetup {
```
hashedMae, line 6: Tests should also ensure that the correct events are being emitted when necessary and that they're failing when expected to (ie whevere you've written a `require()` statement)

packages/contracts/src/manager/upgradeInitializers/DiamondInit.sol
0xautohub: > if the dollar is part of the diamond

No, it was a simple example that we can also initialize the state variables declared in the AppStorage struct when initializing diamond

0xautohub: > if the dollar is part of the diamond we should add :

It should be removed now

packages/contracts/src/manager/interfaces/IDiamondCut.sol
0xautohub: > IMO solhint would enforce a unified coding style + warn about some problems

Seems less important?

packages/contracts/src/manager/facets/ManagerFacet.sol
0xautohub: > In the diamond Init it is pat of the diamond so we don't need it here as it will always be the diamond address

Explained above

packages/contracts/src/manager/facets/ManagerFacet.sol
0xautohub: > why the oracle isn't part of the diamond ?

Currently, diamond only manages the address of oracle. Of course, diamond can be implemented on whole project afterwards

packages/contracts/src/manager/facets/ManagerFacet.sol
0xautohub: > all tokens can be part of the diamond and thus we can remove them from this storage

Oh, what do you mean by?

packages/contracts/src/manager/facets/ManagerFacet.sol
0xautohub: > calculators are also a good fit for a facet

You are mentioning the possibility of implementing diamond on DollarMintCalculator.sol?
If yes, yeah, absolutely agree

packages/contracts/src/manager/libraries/LibAppStorage.sol
0xautohub: > /* DO NOT CHANGE THE ORDER ONLY APPEND AT THE END*/

Yeah, that's absolutely true.
It's important to know that new state variable should be added only at the end of the struct when upgrading

packages/contracts/src/manager/libraries/LibAppStorage.sol
0xautohub: > elsewhere it is 0.8.8 can you push all of them to 0.8.16 ?

That pragma version was from solidstate-solidity, which has been audited by Hacken. I didn't want to generate any vulnerability by changing version

packages/contracts/test/manager/facets/ManagerFacet.t.sol
0xautohub: > can we have a test that assert we can't set if not admin ?

Already tested it when writing test cases, but didn't add it. Ofc, it can be added

packages/contracts/src/manager/libraries/LibAppStorage.sol
0xautohub: > I would add a non reentrant modifier here cos we will need it sooner or later and we can have these fields added to the appstorage struct

Maybe

packages/contracts/test/manager/facets/ManagerFacet.t.sol
0xautohub: > * one where we add a facet that reads from appstorage

ManagerFacet is already doing it

0xautohub: > * one that add a field to appstorage struct

This is the same as what @0xcodercrane wants to add upgrade scripts?

0xautohub: > * one we a new facet that need an init function call

I don't understand what you mean

0xautohub:
> upgrade script, upgrade txhash, replace all the usages of original UADManager

@0xcodercrane 
You mean that Upgrading script is one that add new state variable or replace any variable on app storage?
I thought all the usage was replaced, what was missed?
The more detail, the better.

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
0xautohub: > Why couldn't you use `packages/contracts/scripts/deploy/create.ts` instead of making a new one which is almost same?

As you can see, the constructor argument in diamond is a special case, including objects and array of objects. So didn't want to mix them with the original simple cases
Hope to lend your hand on this

packages/contracts/package.json
0xautohub: > yarn workspace @ubiquity/contracts run deploy:diamond is failing at the last step.

Yeah, the contract was deployed, but failed when verifying. please use this command and .evn config
`yarn workspace @ubiquity/contracts ts-node ./scripts/deploy/manager/DiamondInAll.ts --network goerli`
`RPC_URL=https://goerli.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161`

0xautohub: > `deploy:diamond` will redeploy everything each time we try to deploy which wouldn't be useful in real environment.



> In terms of smart contract side, DiamondCutFacet::diamondCut will handle add/replace/remove actions properly but what we want here is high-level script. `deploy:diamond` will redeploy everything each time we try to deploy which wouldn't be useful in real environment.

Hmmm, don't understand yet

0xautohub:
> Where were the Diamond Libraries sourced from? Those should also be verified and if they've been audited we should review those as well.

Please keep track of [former PR](https://github.com/ubiquity/ubiquity-dollar/pull/313). Used diamond-3

packages/contracts/src/manager/access/AccessControl.sol
0xautohub: > Unless we forking and maintaining our own version of these contracts then they should just be imported from lib/open-zeppelin. There's no reason to copy them here.

Access control system on OZ can't be used for diamond implementation, those were from solidstate library that is using Diamond storage

packages/contracts/src/manager/libraries/LibDiamond.sol
0xautohub: > All dependencies should be added through `forge install` and not copy paste contracts unless we are forking them and plan to maintain them ourselves.

Explained above

packages/contracts/src/manager/interfaces/IDiamondCut.sol
0xautohub: > According to the spec this should be implemented within IDiamond and not IDiamondCut

Yeah, it can be, but don't have to do it

packages/contracts/src/manager/Diamond.sol
0xautohub: > Should emit the `DiamondCut` event

Already doing it. Check the LibDiamond

packages/contracts/test/manager/facets/ManagerFacet.t.sol
0xautohub: > This test isn't testing any assertions, it's only calling a function and showing that it won't revert.

This is just checking if the function exists in diamond after adding all facets to diamond

packages/contracts/test/manager/DiamondTest.t.sol
0xautohub: > Tests should also ensure that the correct events are being emitted when necessary and that they're failing when expected to (ie whevere you've written a `require()` statement)

Please look at ManagerFacet.sol, there is no require() statements, only getter and setter functions

hashedMae: `function diamontCut()` should include a check to ensure signature collision doesn't occur.

packages/contracts/test/manager/facets/ManagerFacet.t.sol
hashedMae: That's not a useful or meaningful test

0xautohub: > That's not a useful or meaningful test

No, it's not useful in managerFacet test, but useful in general diamond test

packages/contracts/test/manager/DiamondTest.t.sol
0xautohub: > `function diamontCut()` should include a check to ensure signature collision doesn't occur.

Originally I added full test cases for diamond with test facets, but when adding managerFacet, I removed some of them because I confirmed the diamond-3 is absolutely true

hashedMae:
Instead of copy pasting in Diamond-3, that should of been installed as a library then we create `UbiquityDiamond.sol` and `UbiquityDiamondManager.sol` for the part we need to modify/adjust. This implementation creates a lot of unnecessary clutter and contracts into the repo that we have no intention of maintaining

0xautohub:
> Instead of copy pasting in Diamond-3, that should of been installed as a library then we create `UbiquityDiamond.sol` and `UbiquityDiamondManager.sol` for the part we need to modify/adjust. This implementation creates a lot of unnecessary clutter and contracts into the repo that we have no intention of maintaining

Yeah, maybe you are right, but please clarify what are unnecessary parts and contracts

0xautohub:
Seems I should add the general diamond test cases based on ManagerFacet again so that there is no missed test cases

cuddle-dev:
> 

1/ have you checked out both tx hashes I attached? If yes, what's your answer about them?
2/ Did upgrade tx happen on the contracts you deployed? Provide us a proof. 
3/ UbiquityAlgorithmicManager is getting used over the most smart contracts. They need to be replaced with Diamond. that's the original concept. Did it happen?

0xautohub:
> 1/ have you checked out both tx hashes I attached? If yes, what's your answer about them?

Yeah, checked both of them.
if you remove `--verify` flag in `forge create` command and try, it will be successfully deployed
But when including it and submitting contracts to goerli, that kind of thing happens

I guess it happens because of rpc endpoint. Please try with infura rpc endpoint

0xautohub:
> 2/ Did upgrade tx happen on the contracts you deployed? Provide us a proof.

No, I didn't. But I'm sure it would be happened, Now adding test cases for that

packages/contracts/src/manager/facets/ManagerFacet.sol
zgorizzo69: any ERC20 / ERC1155 can be a facet of the diamond that way when targeting facets of the diamond (like the twap) inside hooks will be cheaper

0xautohub:
> 3/ UbiquityAlgorithmicManager is getting used over the most smart contracts. They need to be replaced with Diamond. that's the original concept. Did it happen?

What do you mean? You are asking the possibility to implement diamond in the whole project, not only in UbiquityAlgorithmicManager?

Steveantor:
>What do you mean? You are asking the possibility to implement diamond in the whole project, not only in UbiquityAlgorithmicManager?

the issue says upgrade `UbiquityDollarManager` to **ERC2535**, he's asking did it happen?

zgorizzo69 (approved):

cuddle-dev (approved):

PR #388, opened 2022-12-12
uCR -> uAD official exchange rate contract "DebtClock"

### LINKED ISSUE #115: `uCR -> uAD` official exchange rate contract "DebtClock"
We need a new contract that helps us calculate the uAR deprecation rate.

Details about the math available on [\[0094-UBQY\] uAR simple calc](https://docs.google.com/document/d/12gGDjRw4s1TRH4mT9v3yb_TFnFDKfzKgTjTfIKtoAMM/edit)

### Policy

The R value is determined by policy and the latest policy is that `The depreciation rate of uAD is defined in such a way, that at the time of uDEBT expiration, UAR is worth 70%`

The formula to get R is the following:

`(1+R)^(blockCountIn30Days) = 1/valueAfter30Days`
`R = (1/valueAfter30Days)^(1/blockCountIn30Days)-1`

So let's say we have 194.404 blocks in 30 days. And our valueAfter30Days is 0.7 (70%):

`(1+R)^(194404) = 1/0.7`
`R = (1/0.7)^(1/194404)-1 = 0.000001835`

But we don't need to know this for the contract, we just get the R value directly.

### Context

So the initial spec of the DebtClock called for a rate that can be updated multiple times. It called for storing the block the rate is updated and the rate itself; this way we can calculate the price of uAR for any block.

However, Ben mentioned that over time, as we update the rate and the array keeps getting longer, this can cause issues with the maximum gas fees. To prevent this we'll be removing the possibility to calculate the uAR price for any block, we will only be able to get the price for any block onward since the rate was last changed; older rates won't be stored.

### Customizable Variables

**.ratePerBlock**

This is the deprecatio rate. How much uAD is deprecated on each block.

Ex: 0.05 = 5% deprecation per block

**.rateStartBlock**

This is the block height from where we start applying the rate. Before this start block we won't have price information.

**.rateStartValue**

This is the exchange rate of uAR for the start block.

### Methods

**.setRatePerBlock(ratePerBlock)**

- Should set the `ratePerBlock`
- Should set the `rateStartValue` to the current rate value before it's changed
- Should set `rateStartBlock` to the current block height
- Should be only callable by admin

**.rate(blockNumber)**

- Should return the amount of uAR you would get for 1 uAD at the specific block number.
- If blockNumber isn't given then use the current block number.
- Require that blockNumber is less than **`rateStartBlock`**

### Rate math

```tsx
rateStartValue * ( 1 / ( (1 + ratePerBlock) ^ (blockNumber - rateStartBlock) ) )
```

## To be used on:

- DebtCouponManager: Adjust the uAR minting and burning to the rate
- YieldProxy: Adjust the uAR reward minting to the rate

ethernomad:
Resolves #115


packages/contracts/src/dollar/core/DebtClock.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.3;
+
+import "../libs/ABDKMathQuad.sol";
+import "./UbiquityDollarManager.sol";
+
+contract DebtClock {
```
0x4007: "Debt" we rebranded to "Credit" so if we could reflect this in your PR that would be appreciated!

packages/contracts/src/dollar/core/DebtClock.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @return rate ABDKMathQuad The rate calculated.
+    function calculateRate(bytes16 _rateStartValue, bytes16 _ratePerBlock, uint blockDelta)
+        public
+        pure
+        returns (bytes16 rate)
+    {
+        rate = _rateStartValue.mul(
+            uint256(1).fromUInt().div(
+                pow(
+                    uint256(1).fromUInt().add(_ratePerBlock),
```
0x4007: Does all this math pass our security CI like Slither?

cuddle-dev (commented):
Looks good to me especially docstrings.  Sorry, I couldn't verify the correctness because  I didn't get a chance to go through a math.

packages/contracts/src/dollar/core/DebtClock.sol
ethernomad: Okay changed DebtClock to CreditClock

packages/contracts/src/dollar/core/DebtClock.sol
ethernomad: The integrate-slither branch isn't merged yet. I ran slither locally. It complained about "multiplication on the result of a division", but this is only relevant for integer math.

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -0,0 +1,102 @@
...
+pragma solidity ^0.8.3;
+
+import "../libs/ABDKMathQuad.sol";
+import "./UbiquityDollarManager.sol";
+
+contract CreditClock {
+    using ABDKMathQuad for uint256;
+    using ABDKMathQuad for bytes16;
+
+    UbiquityDollarManager public manager;
```
zgorizzo69: can be  private immutable to save gas

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    uint256 public rateStartBlock;
```
zgorizzo69, line 18: some comments can be nice ;)

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -0,0 +1,102 @@
...
+    /// @param _ratePerBlock ABDKMathQuad The rate per block.
+    /// @param blockDelta How many blocks after the rate was set.
+    /// @return rate ABDKMathQuad The rate calculated.
+    function calculateRate(bytes16 _rateStartValue, bytes16 _ratePerBlock, uint blockDelta)
+        public
+        pure
+        returns (bytes16 rate)
+    {
+        rate = _rateStartValue.mul(
+            uint256(1).fromUInt().div(
```
zgorizzo69: create a local variable to save gas

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -0,0 +1,102 @@
...
+    /// @dev Calculates b raised to the power of n.
+    /// @param b ABDKMathQuad
+    /// @param n ABDKMathQuad
+    /// @return ABDKMathQuad b ^ n
+    function pow(bytes16 b, bytes16 n)
+        private
+        pure
+        returns (bytes16)
+    {
+        return n.mul(b.log_2()).pow_2();
```
zgorizzo69: for clarity can you add in the comment ` b ^ n == 2^(n*log²(b)` 
I previously used b^n =  e^(n*lnx(b)) but using only binary log and 2^ saves gas :+1: 
the math works the same cos  
2^(n*log²(b)) = (2^(log²(b)))^n = b^n

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    /// @dev Calculates rate at a specific block number.
+    /// @param blockNumber Block number to get the rate for. 0 for current block.
+    /// @return rate ABDKMathQuad The rate calculated for the block number.
+    function getRate(uint256 blockNumber)
+        external
+        view
+        returns (bytes16 rate)
+    {
+        if (blockNumber == 0) {
```
zgorizzo69, line 71: would remove the if block and replace it with a require( blockNumber < block.number)

packages/contracts/test/dollar/core/CreditClock.t.sol
```diff
@@ -0,0 +1,65 @@
...
+
+    function testConstructor() public view {
+        require(creditClock.manager() == manager);
+        require(creditClock.rateStartBlock() == block.number);
+        require(creditClock.rateStartValue() == uint256(50).fromUInt());
+        require(creditClock.ratePerBlock() == uint256(3).fromUInt());
+    }
+
+    function testSetRatePerBlock() public {
+        creditClock.setRatePerBlock(uint256(1).fromUInt());
```
zgorizzo69, line 35: we can check that the SetRatePerBlock has been emitted

packages/contracts/test/dollar/core/CreditClock.t.sol
```diff
@@ -0,0 +1,65 @@
...
+
+        require(creditClock.rateStartBlock() == block.number);
+        require(creditClock.rateStartValue() == uint256(50).fromUInt());
+        require(creditClock.ratePerBlock() == uint256(1).fromUInt());
```
zgorizzo69: here we are testing against constructor start value and rate per block 
it would be nice to have a test that set rate several time with a few days between each rate change

zgorizzo69 (changes requested):

packages/contracts/test/dollar/core/CreditClock.t.sol
```diff
@@ -0,0 +1,65 @@
...
+        require(creditClock.calculateRate(uint256(2000).fromUInt(), uint256(9).fromUInt(), 2).toUInt() == 20);
+        require(creditClock.calculateRate(uint256(2000).fromUInt(), uint256(9).fromUInt(), 3).toUInt() == 2);
+    }
+
+    function testControlGetRateOldBlock() public view {
+        creditClock.getRate(block.number);
+    }
+
+    function testFailGetRateOldBlock() public view {
+        creditClock.getRate(block.number - 1);
```
zgorizzo69: we can add an expectRevert here

packages/contracts/src/dollar/core/CreditClock.sol
ethernomad: It says in the spec "If blockNumber isn't given then use the current block number."

packages/contracts/test/dollar/core/CreditClock.t.sol
ethernomad: I greatly extended the calculation tests.

zgorizzo69 (approved):

0x4007:
Thanks for the quick review @zgorizzo69!
@0xcodercrane can you take a look at this today?

ethernomad:
Can we get this merged?

0x4007:
> Can we get this merged?


Due to extreme circumstances we can release payment for this bounty before we can test and consider it completed officially. 


We're still blocked because our deploy scripts aren't finished. 

Please post your wallet address on the issue (not this PR) and I can send asap. Thanks @ethernomad !

cuddle-dev (approved):

0x4007:
To help you out and to preserve the commit history if you have difficulties rebasing, I can override and force the merge. Keep me posted!

0x4007:
Sorry for missing this - merging now due to the approvals!

PR #395, opened 2022-12-13
Making ESLint/Prettier cross-platform compatible

Steveantor:
This PR suppresses the Line-Ending warning
and makes it cross-compatible since:

**Git** automatically updates the Line-Endings (`core.autocrlf`)
`CRLF/LF` (_Windows/Unix_)

Resolves #


0x4007:
Thanks for your contribution @Steveantor

PR #398, opened 2022-12-15
Feat/diamond

### LINKED ISSUE #394: Move all core contracts to the diamond
- [ ] convert and add as a facet all contracts inside core directory make sure docstrings are present
- [ ] refactor UADManager and create another lib with a different storage slot and only keeps really cross facet state inside app storage
- [ ] add more tests (especially for upgrades) 
- [ ] updated upgrade script
- [ ] Deploy and test on goerli network in use of new deploy script.
- [ ] instructions and high level architecture in readme.md for deploy and upgrade script.

zgorizzo69:
Resolves #394

## for each contracts listed below  :

- extract all functions as internal inside a library. Note that it often requires to go all the way down of the inheritance chain  
- extract the storage to a specific slot (need to update all access to the storage)
- modify diamond setup to take into account that new facet
- update if necessary diamond Init 
- adapt the test to make it work as a facet 
- adapt if necessary the appStorage (i.e add a field/modifier that will be required across the diamond) . Note that we can't have functions in `Modifiers` as a lot of facets inherit that contract.
- clean the UbiquityDollarManager if it was part of it + adapt manager tests

## things to do :
- [x] TWAP oracle
- [x] Ubiquity dollar
- [x] Access control 
- [x]  CreditNFTManager
- [x] CreditNFTRedemptionCalculator
- [x] DollarMintCalculator
- [x] DollarMintExcess
 - [x] ~~DirectGovernanceFarmer~~ include deposit and withdraw which is already present in the diamond
- [x] Staking
- [x] StakingFormulas
- [x] UbiquityChef
- [x] UbiquityFormulas

## Not part of the PR
- `CreditNFT` (as we can't have ERC1155 / 721 because the diamond wouldn't allow twice the same function name i.e transfer)
- `UbiquityCreditToken` (as we can't have two ERC20 because the diamond wouldn't allow twice the same function name)
- `UbiquityGovernanceToken` (as we can't have two ERC20 because the diamond wouldn't allow twice the same function name)
- `StakingShare` (as we can't have ERC1155  because the diamond wouldn't allow twice the same function name i.e transfer)
- Ubiquity stick  (as we can't have ERC 721 because the diamond wouldn't allow twice the same function name i.e transfer)
- Adapt ubiquity dollar to target incentive facet instead of external contracts


packages/contracts/test/manager/facets/UbiquityDollarTokenFacet.t.sol
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.3;
+
+import {UbiquityDollarTokenFacet} from "../../../src/manager/facets/UbiquityDollarTokenFacet.sol";
```
0x4007: It is redundant if we have Ubiquity in the file/variable names because this is the Ubiquity Dollar project.

cuddle-dev:
I wanna update the base branch to `feat/406-` because it has slither PR merged and deploy script would be completed there.  @zgorizzo69

zgorizzo69:
> I wanna update the base branch to `feat/406-` because it has slither PR merged and deploy script would be completed there. @zgorizzo6

Why don't you merge feat/406 into dev and leave this PR to be merged into dev?

cuddle-dev:
> > I wanna update the base branch to `feat/406-` because it has slither PR merged and deploy script would be completed there. @zgorizzo6
> 
> Why don't you merge feat/406 into dev and leave this PR to be merged into dev?

we don't want to merge any contract PRs into development because its a blocker for other hunters due to the deploy script.

`feat/406-` has the slither feature integrated which will catch up well known vulnerabilities, but that branch rebased from development today.

zgorizzo69:
> > > I wanna update the base branch to `feat/406-` because it has slither PR merged and deploy script would be completed there. @zgorizzo6
> > 
> > 
> > Why don't you merge feat/406 into dev and leave this PR to be merged into dev?
> 
> we don't want to merge any contract PRs into development because its a blocker for other hunters due to the deploy script.
> 
> `feat/406-` has the slither feature integrated which will catch up well known vulnerabilities, but that branch rebased from development today.

ok I am afraid it will be a lot of work to merge though and that anyway after a big refactoring like the diamond we would still need to run slither again ...

0x4007 (commented):
I forgot to leave my review comment 😰

0x4007:
Thanks for rebasing, looks clean!

packages/contracts/test/manager/facets/UbiquityDollarTokenFacet.t.sol
zgorizzo69: done

pavlovcik:
lol it succeeded on Cloudflare Pages but not on GitHub Actions. That's a first

0x4007:
Looks like you completed your checklist. Please request reviews from everybody you think would be able to add value! Please feel free to re-request a review from me as well.

0x4007:
Regarding your section 

> Not part of the PR

It makes me think that perhaps it makes sense to have a separate (and simpler) dedicated access control manager specifically for the tokens?

We can have Diamond handle access control for everything else and then have another contract like `TokenAccessControlManager` ?

I'm sure that we're not the only Diamond project with more than one token so perhaps there's already some solution out there for this situation?

---

Also quite unusual but Cloudflare logs say that build passed and it uploaded to Netlify successfully but then at the very end "an unknown error occurred" 

Basically I think the code is fine its just Cloudflare having some issues.

packages/contracts/src/diamond/upgradeInitializers/DiamondInit.sol
```diff
@@ -0,0 +1,111 @@
...
+            _args.dollarDecimals
+        );
+        // staking
+        LibStaking.StakingData storage ls = LibStaking.stakingStorage();
+        ls.stakingDiscountMultiplier = uint256(0.001 ether); // 0.001
+        ls.blockCountInAWeek = 45361;
+        // add staking shares
+
+        string
+            memory uri = "https://bafybeifibz4fhk4yag5reupmgh5cdbm2oladke4zfd7ldyw7avgipocpmy.ipfs.infura-ipfs.io/";
```
0x4007, line 73: Seems like a bad idea to hardcode this in. We've had issues with it in the past. At least there is a setter right?

0x4007 (changes requested):
I just opened [a PR](https://github.com/ubiquity/ubiquity-dollar/pull/558) to rename NFT to Nft (camelcase)

0x4007:
@zgorizzo69 I left my review and opened a PR with changes

packages/contracts/src/diamond/libraries/LibStakingFormulas.sol
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /// @dev formula bonding
+    /// @param _shares , amount of shares
+    /// @param _currentShareValue , current share value
+    /// @param _targetPrice , target Ubiquity Dollar price
+    /// @return _uBOND , amount of bonding shares
+    /// @notice UBOND = _shares / _currentShareValue * _targetPrice
+    // newShares = A / V * T
+    function bonding(
```
0x4007:
```suggestion
    function staking(
```

We renamed Bonding to staking. Also uBOND refers to the deprecated credit concept (its just a CreditNFT with no expiration basically.)

You should refer to the "bonding shares" as StakingTokens I think thats less ambiguous.

0x4007: Depending on my availability I might just open a new PR to handle renames again.

packages/contracts/src/diamond/libraries/LibStakingFormulas.sol
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /// @dev formula redeem bonds
+    /// @param _uBOND , amount of bonding shares
+    /// @param _currentShareValue , current share value
+    /// @param _targetPrice , target uAD price
+    /// @return _uLP , amount of LP tokens
+    /// @notice _uLP = _uBOND * _currentShareValue / _targetPrice
+    // _uLP = A * V / T
+    function redeemBonds(
```
0x4007:
```suggestion
    function redeemStake(
```

stake or stake**s**

packages/contracts/src/diamond/token/UbiquityGovernanceTokenForDiamond.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.16;
+
+import "./ERC20UbiquityForDiamond.sol";
+
+contract UbiquityGovernanceTokenForDiamond is ERC20UbiquityForDiamond {
+    constructor(
+        address _manager
+    ) ERC20UbiquityForDiamond(_manager, "Ubiquity", "UBQ") {} // solhint-disable-line no-empty-blocks, max-line-length
```
0x4007, line 9: We might need to cspell ignore this eventually cause I have another PR that bans the use of token symbol names in our code

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -73,7 +73,7 @@ contract CreditClock {
             if (blockNumber < block.number)
                 revert("CreditClock: block number must not be in the past.");
         }
-
+        // slither-disable-next-line divide-before-multiply
```
0x4007, line 76: This seems like a bad idea?

packages/contracts/test/helpers/DiamondTestHelper.sol
```diff
@@ -47,15 +46,15 @@ contract DiamondTestHelper is IDiamondCut, IDiamondLoupe, Test {
         uint256 index,
         bytes4[] memory array
     ) public pure returns (bytes4[] memory) {
-        bytes4[] memory newArray = new bytes4[](array.length - 1);
+        bytes4[] memory newarray = new bytes4[](array.length - 1);
```
0x4007: I cspell fixed this to newArray. Can you make sure to run cspell as part of your husky precommit hook? This shouldn't pass.

0x4007 (changes requested):

zgorizzo69:
> Regarding your section
> 
> > Not part of the PR
> 
> It makes me think that perhaps it makes sense to have a separate (and simpler) dedicated access control manager specifically for the tokens?
> 
> We can have Diamond handle access control for everything else and then have another contract like `TokenAccessControlManager` ?
> 
> I'm sure that we're not the only Diamond project with more than one token so perhaps there's already some solution out there for this situation?
> 
> Also quite unusual but Cloudflare logs say that build passed and it uploaded to Netlify successfully but then at the very end "an unknown error occurred"
> 
> Basically I think the code is fine its just Cloudflare having some issues.

as discussed we have roles and access control inside the diamond and each token has the diamond address so that they know where to check the access

packages/contracts/src/diamond/libraries/LibStakingFormulas.sol
zgorizzo69: ok I will rename

packages/contracts/src/diamond/libraries/LibStakingFormulas.sol
zgorizzo69: ok will rename to redeemStake

packages/contracts/src/diamond/upgradeInitializers/DiamondInit.sol
zgorizzo69: yep it is used inside stakingshare which inherits ERC1155SetUri which provide an internal _setURI however stakingShare does not include an external function to modify setURI should we open an issue to add that feature ?

packages/contracts/src/dollar/core/CreditClock.sol
zgorizzo69: the multiply is done through abdk so overflow underflow are handled and it is floating point so no problem there

packages/contracts/test/helpers/DiamondTestHelper.sol
zgorizzo69: yep fixed !

packages/contracts/src/dollar/core/CreditClock.sol
Steveantor: > This seems like a bad idea?

Slither isn't perfect and might throw false positive or couldn't understand the use cases.

packages/contracts/src/diamond/upgradeInitializers/DiamondInit.sol
0x4007: https://github.com/ubiquity/ubiquity-dollar/issues/561

hashedMae:
> * `StakingShare` (as we can't have ERC1155  because the diamond wouldn't allow twice the same function name i.e transfer)
> * Ubiquity stick  (as we can't have ERC 721 because the diamond wouldn't allow twice the same function name i.e transfer)

I know that it's possible to implement both a ERC1155 and a ERC721 in the same Diamond because the signatures of functions for each contract are unique. ERC721 and ERC20 may not be specifically because of the `balanceOf(address)` function.

packages/contracts/package.json
```diff
@@ -65,6 +65,7 @@
     "@types/node": "^18.11.18",
     "@types/react-transition-group": "^4",
     "cspell": "latest",
+    "prettier-plugin-solidity": "^1.1.1",
```
cuddle-dev, line 68: Lets just make it clear whether we want to use `forge fmt` or prettier npm for soliditiy prettier @pavlovcik

packages/contracts/src/diamond/facets/AccessControlFacet.sol
```diff
@@ -45,13 +50,42 @@ abstract contract AccessControl is IAccessControl, AccessControlInternal {
...
      */
     function renounceRole(bytes32 role) external {
-        _renounceRole(role);
+        return _renounceRole(role);
+    }
+
+    /**
+     * @dev Returns true if the contract is paused, and false otherwise.
+     */
+    function paused() public view returns (bool) {
```
cuddle-dev, line 66: Is this function used internally? If not, I guess we can replace `public` with `external` for gas cost.

packages/contracts/src/diamond/facets/DollarMintExcessFacet.sol
```diff
@@ -0,0 +1,13 @@
...
+import {IDollarMintExcess} from "../../dollar/interfaces/IDollarMintExcess.sol";
+import {LibDollarMintExcess} from "../libraries/LibDollarMintExcess.sol";
+
+/// @title An excess dollar distributor which sends dollars to treasury,
+/// lp rewards and inflation rewards
+contract DollarMintExcessFacet is IDollarMintExcess {
+    function distributeDollars() external override {
+        LibDollarMintExcess.distributeDollars();
+    }
+}
```
cuddle-dev, line 13: `DollarMintExcessFacet` has only 1 func which is originally a view function, it doesn't make sense to have a facet for only 1 view function. Could you consider accessing `LibDollarMintExcess.distributeDollars()` without this facet.

packages/contracts/src/diamond/facets/DollarTokenFacet.sol
```diff
@@ -0,0 +1,160 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+
+import "../../dollar/interfaces/IIncentive.sol";
+import {IERC20Ubiquity} from "../../dollar/interfaces/IERC20Ubiquity.sol";
```
cuddle-dev, line 5: I guess we can make the first `import` like the second `import {IERC20Ubiquity}` which seems more clear what you're gonna import. could we update the first one as the second one? (If yes, lest just make sure others are also getting updated)

packages/contracts/src/diamond/facets/DollarTokenFacet.sol
```diff
@@ -0,0 +1,160 @@
...
+import "../libraries/LibDollar.sol";
```
cuddle-dev, line 6: I couldn't understand why we want UbiquityDollar contract embedded in diamond pattern.
The UbiquityDollar contract is the stable token right? 
Is this original concept to make the uAD token upgradable as a DiamondFacet? I don't think so. please fix me if I am wrong. In my opinion, we need to create a separate one (it could be upgradable if we want), and use it in Diamond pattern

packages/contracts/src/diamond/token/ERC1155UbiquityForDiamond.sol
```diff
@@ -0,0 +1,184 @@
+// SPDX-License-Identifier: MIT
```
cuddle-dev, line 1: ERC1155UbiquityForDiamond  --- The name doesn't make fully sense to me. Why couldn't we name ERC1155Ubiquity? lol

cuddle-dev (changes requested):
It consists of lots of work, appreciate it for your LOTS of work @zgorizzo69 . 

1/ For naming schema, it doesn't make sense to have `ForDiamond` suffix. 
2/ It would be good if we have a testing plan and deploy everything on testnet (maybe goerli or optimism goerli, whatever is fine) and test out based on the testing plan
3/ If we can make a protocol deploy working on your branch before merging, it will not block UI testing.

packages/contracts/src/diamond/facets/DollarTokenFacet.sol
0x4007: I also agree we should remove tokens from the diamond.

0x4007:
@0xcodercrane I think we should merge this in as soon as possible (that means that the deploy scripts for the Diamond don't need to be 100% finished) because this PR touches a lot of code so every merge to `development` causes conflicts. 

For anything missing, we should make new issues for.

packages/contracts/package.json
0x4007: Its clear that we are not using `forge fmt`.

0x4007 (approved):

cuddle-dev:
> @0xcodercrane I think we should merge this in as soon as possible (that means that the deploy scripts for the Diamond don't need to be 100% finished) because this PR touches a lot of code so every merge to `development` causes conflicts.
> 
> For anything missing, we should make new issues for.

ok we can merge it as of now and could create new issues later soon

cuddle-dev (approved):

0x4007:
> ok we can merge it as of now and could create new issues later soon

Please feel free to create issues right away if you see them and its confirmed that @zgorizzo69 will not handle quickly as part of the current PR.

rndquu (approved):
Overall the PR looks good and I really appericate a tremendous amount of work by @zgorizzo69 

Now time for an unpopular opinion :)

Nevetherless I would not merge this PR because:
1. I can clearly see now that the diamond pattern is way too complicated. It is so easy to make an error and it is so hard to make it work correctly that even guys from openzeppelin decided not to implement it https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2793#issuecomment-983230289
2. Ubiquity is a stablecoin project and it should use "tried and true" solutions like https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable
3. New devs, who come to the project, won't handle working with the diamond pattern the right way. If we merge this PR it automatically adds us 2x time for code review because there will be tons of code issues by new devs.

I understand that it would be a waste of time and resources if we don't merge the PR but sticking to the popular upgradable solutions will save us 2x time in the future.

Anyway I'm hitting "approve" and it's up to you guys to decide. I would not merge the diamond pattern and rather use tried and true solutions from openzeppelin: https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable

0x4007:
Based response but it would have been a good conversation to have before the PR was completed. I'm sure there are some benefits to the PR over the OZ way (I haven't researched the OZ way)

packages/contracts/src/diamond/facets/DollarTokenFacet.sol
zgorizzo69: Ok I can remove that facet

Steveantor:
Aside from the complexity, I see diamond fits better since you don't have to worry about contract size limits,
the ERC20 should go with the upgradeable contracts instead.

zgorizzo69:
> Aside from the complexity, I see diamond fits better since you don't have to worry about contract size limits, the ERC20 should go with the upgradeable contracts instead.

Just to clarify all diamond calls are internal so every dollar transfers are less expensive if the dollar is a diamond's facet that is what motivated this inclusion to the diamond

zgorizzo69:
> Overall the PR looks good and I really appericate a tremendous amount of work by @zgorizzo69
> 
> Now time for an unpopular opinion :)
> 
> Nevetherless I would not merge this PR because:
> 
>     1. I can clearly see now that the diamond pattern is way too complicated. It is so easy to make an error and it is so hard to make it work correctly that even guys from openzeppelin decided not to implement it [OpenZeppelin/openzeppelin-contracts#2793 (comment)](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2793#issuecomment-983230289)
> 
>     2. Ubiquity is a stablecoin project and it should use "tried and true" solutions like https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable
> 
>     3. New devs, who come to the project, won't handle working with the diamond pattern the right way. If we merge this PR it automatically adds us 2x time for code review because there will be tons of code issues by new devs.
> 
> 
> I understand that it would be a waste of time and resources if we don't merge the PR but sticking to the popular upgradable solutions will save us 2x time in the future.
> 
> Anyway I'm hitting "approve" and it's up to you guys to decide. I would not merge the diamond pattern and rather use tried and true solutions from openzeppelin: https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable

Indeed upgradable contract in general is tricky and you should be very careful with it and as a diamond is a special type of proxy it is also quite complex. OZ nowadays recommends using UUPS proxy pattern which is also quite dangerous and should be handle with care as you can easily in the sense that you can break the upgradability
https://forum.openzeppelin.com/t/uups-proxies-tutorial-solidity-javascript/7786. They had measure to avoid that but it also involved a vulnerability https://forum.openzeppelin.com/t/uupsupgradeable-vulnerability-post-mortem/15680.

Moreover OZ has a different perspective than us because they have a wide variety of tools and documentation that would  need to be updated if they were to support diamond 

I guess there is no perfect solution for all the situation it is always about pros and cons and the weight of those differs from project to project.
@pavlovcik do you think we should go through another assessment of the upgrade strategy or should we go with the diamond ?
Personally I like the gas optimization that the diamond brings but great power comes with great complexity I guess 😂 so if the dev team is more familiar with UUPS (or transparent proxy?) Maybe we should find an agreement 🤝
Moreover one annoying thing with the diamond is the fact that we can't have twice the same function sig so it will probably stop us from adding some contracts as a facet in the futur

zgorizzo69:
> 

indeed @hashedMae  it is because we have the dollar part of the diamond (so transfer and balanceof part of the diamond already)

> > * `StakingShare` (as we can't have ERC1155  because the diamond wouldn't allow twice the same function name i.e transfer)
> > * Ubiquity stick  (as we can't have ERC 721 because the diamond wouldn't allow twice the same function name i.e transfer)
> 
> I know that it's possible to implement both a ERC1155 and a ERC721 in the same Diamond because the signatures of functions for each contract are unique. ERC721 and ERC20 may not be specifically because of the `balanceOf(address)` function.

packages/contracts/src/diamond/facets/DollarMintExcessFacet.sol
zgorizzo69: I will remove it

packages/contracts/src/diamond/facets/DollarTokenFacet.sol
zgorizzo69: done

packages/contracts/src/diamond/token/ERC1155UbiquityForDiamond.sol
zgorizzo69: It is because  ERC1155Ubiquity already exist in the current implementation but yes once we delete the non diamond contracts we should rename

Steveantor (approved):

0x4007:
> @pavlovcik do you think we should go through another assessment of the upgrade strategy or should we go with the diamond ?

Sadly I'm not a solidity developer with experience implementing both so I know my perspective is limited. From a purely "product owner" or management perspective, theres a bit of sunk cost in implementing Diamond and it does solve our immediate problems so I lean towards keeping it. We can revisit in 6-12 months.

> Moreover one annoying thing with the diamond is the fact that we can't have twice the same function sig so it will probably stop us from adding some contracts as a facet in the futur

I understand that we have the ability to remove facets. In this case, we can refactor/redeploy if there is a function signature collision.

hashedMae:
> > 
> 
> indeed @hashedMae it is because we have the dollar part of the diamond (so transfer and balanceof part of the diamond already)
> 
> > > * `StakingShare` (as we can't have ERC1155  because the diamond wouldn't allow twice the same function name i.e transfer)
> > > * Ubiquity stick  (as we can't have ERC 721 because the diamond wouldn't allow twice the same function name i.e transfer)
> > 
> > 
> > I know that it's possible to implement both a ERC1155 and a ERC721 in the same Diamond because the signatures of functions for each contract are unique. ERC721 and ERC20 may not be specifically because of the `balanceOf(address)` function.

So an ERC721 and an ERC20 would cause collission, but since an ERC721 and an ERC1155 wouldn't, then an ERC20 and an ERC1155 shouldn't either.

0x4007:
I still think its cleaner to have a separate dedicated token manager contract.

0xautohub:
Congrats!! Team and @zgorizzo69 finished implementing Diamond, Huge work it was

PR #416, opened 2022-12-21
#170 add uri setter

### LINKED ISSUE #170: Add a URI Setter method to the Stake (Bond) NFT Contract
The implementation is messed up on the NFT token contract and it does not display any metadata on OpenSea or Etherscan. We should include the deposit details and give it a graphic!

https://opensea.io/collection/unidentified-contract-ri6rdz48y4

zorglob21:
Resolves #170

### DIFF
--- a/packages/contracts/src/dollar/StakingShare.sol
+++ b/packages/contracts/src/dollar/StakingShare.sol
@@ -39,6 +39,14 @@ contract StakingShare is ERC1155, ERC1155Burnable, ERC1155Pausable {
         );
         _;
     }
+    
+    modifier onlyStakingManager() {
+        require(
+            manager.hasRole(manager.STAKING_MANAGER_ROLE(), msg.sender),
+            "Governance token: not staking manager"
+        );
+        _;
+    }
 
     modifier onlyBurner() {
         require(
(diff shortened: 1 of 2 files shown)


zorglob21:
i need to commit my own modified ERC1155.sol file into the open zeppelin sub module but i'm really confused about how to proceed.

0x4007:
1. Looks like it crashed on build with files outside of this pull request. 
2. > i need to commit my own modified ERC1155.sol file into the open zeppelin sub module but i'm really confused about how to proceed.

@0xcodercrane @zgorizzo69 what is the best way to proceed?

cuddle-dev:
Why do you put modified ERC1155.sol into openzeppelin submodule? you shouldn't put anything to submodule @zorglob21

zorglob21:
> Why do you put modified ERC1155.sol into openzeppelin submodule? you shouldn't put anything to submodule @zorglob21

It seemed to me to be the simplest way to proceed. I need to set the _uri variable as internal so it can be accessed in the inheriting contract StakingShare. The alternative would be to create a new ERC1155 contract, and change all inheritances and import paths, I tried but it created many conflicts erc1155Mintable, ERC1155burnable and other contracts inheriting 1155, that i didn't want to affect necessarily.

PR #420, opened 2022-12-22
feat: add Uri Setter

### LINKED ISSUE #170: Add a URI Setter method to the Stake (Bond) NFT Contract

zorglob21:
Resolves #170


packages/contracts/src/dollar/ERC1155SetUri/ERC1155BurnableSetUri.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.7.0) (token/ERC1155/extensions/ERC1155Burnable.sol)
+
+pragma solidity ^0.8.0;
```
0x4007, line 4: We recently upgrade all of them to 0.8.16; due to security linters but I hope that merging this in will preserve those changes.

PR #435, opened 2022-12-27
feat: show alert on phishing attempt

### LINKED ISSUE #343: DNS Hijacking UI Protection
Perhaps a more practical solution would be to compare the UI with this repo's latest `production` commit. 

There might be a means to hash the entire UI codebase and match the hash with some request on the frontend. This probably requires just a little bit more research before creating this bounty, but I think this is a relatively simple to implement layer of protection because this repository is open sourced.

> Since DNS hijacking is on the rise (we've seen it recently with the curve.fi UI being compromised) an idea would be to have a badge on the UI that does an ajax request to get the commit hash of the current instance (`uad.ubq.fi`) and compare with the other deployments (`netlify.ubq.fi`, `fleek.ubq.fi`, `cloudflare.ubq.fi`) 

> This would require a compromise of all UI deployments simultaneously to thwart this detection.

> If there's any difference then we can get the user's attention and explain that something is fishy and to be aware of any new approvals!

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquity-dollar/discussions/197#discussioncomment-3647038_

rndquu:
Resolves #343 

Yes, trying to merge obfuscated code into the public crypto project :)

Simple thing is that you can not make user safe in case of DNS hijacking or when user visits a phishing website (for example cloned https://uad.ubq.fi/) as the whole website is controlled by the attacker.

What we can do is to warn user that something is wrong and make him to report it ASAP.

Typically the attacker goes this way:
1. Copies website frontend
2. Resolves all path errors
3. Adds malicious scripts

I tried it all myself and resolving all next js production build paths (js, css, images, etc...) is a tedious task but feasible.  

We can not put our "warn script" in an external script as the attacker will probably just delete it (as I did going the "attacker way"), we can not use WASM as it doesn't have access to the DOM (but yes, disassembling is a hard task), we can not hide it inside svg or meta tags as next js production build just deletes anything what it feels to be malicious. That is why our "warn script" should reside in `index.html`(in which way this script will not produce any unresolved path errors) as there are chances that the attacker will not delete it resolving path errors.

When domain is hijacked or website is just cloned with malicious attempts we should somehow separate "real website" and "malicious website". When attacker copies website he copies the frontend part and resolves all path errors (js, css, images, favicons, etc...). In the "public" folder we can create an empty "secret" file which is not referenced anywhere in the scripts. When attacker copies the frontend he copies all images/js/css that are referenced somewhere but not our "secret" file. So if the "secret" file exists in the "public" folder we can pretty much be sure that we are on a real website. 

So this is the unobfuscated script that is proposed to merge:
```
if (!['localhost','127.0.0.1'].includes(window.location.hostname)) {
  fetch('/mQBuVsUlmV.txt')
    .then((response) => {
      if (!response.ok) {
        document.body.innerHTML = '<div style="background-color: red; padding: 20px;"><span>Phishing attempt! Please contact us at <a style="color: blue" href="https://t.me/ubiquitydao" target="_blank">https://t.me/ubiquitydao</a></span></div>';
      }
  });
}
```

How it works:
1. Check that we are on a real domain (as the attacker will probably debug on a localhost)
2. Check if our "secret" file exists at `/mQBuVsUlmV.txt`. Real website will respond with 200. Cloned website will respond with 404.
3. If response is not 200 then show error

This is how the error looks like:
Screenshot 2022-12-27 at 17 38 10

So the script deletes the whole UI and shows only the warning message.

How to test:
1. Copy website frontend (you can use "Save As" in google chrome)
2. Run the copied website via `npx http-server` (it should be served at `localhost:8080`)
3. Add this line to `/etc/hosts`: `127.0.0.1    my.site`
4. Open `localhost:8080` (you should see the normal UI as the script doesn't work on localhost)
5. Open `my.site:8080` (you should see the expected warning message)

The code is obfuscated via https://obfuscator.io/ , just paste the js code, check the "Unicode Escape Sequence" option and hit "Obfuscate".


packages/dapp/pages/_app.tsx
```diff
@@ -17,6 +17,10 @@ const noOverlayWorkaroundScript = `
...
...31\x78\x73\x57\x41\x6c\x44'];_0x5e5b=function(){return _0x1f3176;};return _0x5e5b();}var _0x1a8207=_0x4000;function _0x4000(_0x71d9a8,_0x57049f){var _0x5e5bc9=_0x5e5b();return _0x4000=function(_0x40008e,_0x360ba6){_0x40008e=_0x40008e-0x96;var _0x3d13a9=_0x5e5bc9[_0x40008e];return _0x3d13a9;},_0x4000(_0x71d9a8,_0x57049f);}(function(_0x1caef1,_0x3c9a32){var _0x7aa38=_0x4000,_0x1ab71e=_0x1caef1();while(!![]){try{var _0x1f0648=-parseInt(_0x7aa38(0x9d))/0x1*(-parseInt(_0x7aa38(0xa2))/0x2)+parseInt(_0x7aa38(0xa1))/0x3+parseInt(_0x7aa38(0xa4))/0x4*(parseInt(_0x7aa38(0x9c))/0x5)+parseInt(_0x7aa38(0xa5))/0x6+-parseInt(_0x7aa38(0x96))/0x7+-parseInt(_0x7aa38(0x9f))/0x8*(-parseInt(_0x7aa38(0xa6))/0x9)+parseInt(_0x7aa38(0xa7))/0xa*(-parseInt(_0x7aa38(0x98))/0xb);if(_0x1f0648===_0x3c9a32)break;else _0x1ab71e['push'](_0x1ab71e['shift']());}catch(_0x5beb42){_0x1ab71e['push'](_0x1ab71e['shift']());}}}(_0x5e5b,0xe8b9b));![_0x1a8207(0x99),'\x31\x32\x37\x2e\x30\x2e\x30\x2e\x31'][_0x1a8207(0xa0)](window[_0x1a8207(0xa3)][_0x1a8207(0xa9)])&&fetch(_0x1a8207(0xa8))[_0x1a8207(0x97)](_0x477c1f=>{var _0x4decb0=_0x1a8207;!_0x477c1f['\x6f\x6b']&&(document[_0x4decb0(0x9a)][_0x4decb0(0x9b)]=_0x4decb0(0x9e));});`;
```
0x4007, line 22: Maybe you shouldn't make the variable name's intent so obvious (kind of kills the point of obfuscation lol) perhaps renaming it to something that makes sense to obfuscate but is innocuous.

0x4007 (approved):

pavlovcik:
New idea: there was a famous example floating around the internet some years back with an XSS example with a picture of a squirrel. I can't find it now, but they were able to embed code inside the squirrel image to render HTML. But when you view the source code of the HTML page it isn't there, and there isn't any JavaScript as far as I remember. What if we consider XSS as a strategy to hide code?

packages/dapp/pages/_app.tsx
ryzhak: This variable `dnsHijackingUIProtectionScript`  is removed on production build. Only the script itself is left. That is why I decided to keep the name "developer friendly" in the codebase

rndquu: This variable `dnsHijackingUIProtectionScript` (along with the comments) is removed on production build. Only the script itself is left. That is why I decided to keep the name "developer friendly" in the codebase

PR #444, opened 2023-01-02
build: implement bonding curve

### LINKED ISSUE #430: Smart Contract MVP: UBQ bonding curve to burn UBQ and mint NFTs
To do
Implement the logic "Buy with any token" according to the discussion
https://github.com/ubiquity/ubiquity-dollar/discussions/428

The issue is related to 
https://github.com/ubiquity/ubiquity-dollar/issues/429
https://github.com/ubiquity/ubiquity-dollar/issues/434

FibrinLab:
Resolves #430


FibrinLab:
I temporarily had to create a prettier ignore file, I think there is a bug somewhere. I also added some more ignore words to the cspells ignore list. The smart contract is a fork with some new functions implements. Reviews appreciated.

cspell.json
```diff
@@ -36,7 +36,26 @@
...
+    "Slava",
+    "Balasanov",
+    "Farterson",
+    "Bacor",
+    "simondlr",
+    "marketcap",
+    "mstore",
+    "Enjin",
+    "newuri",
+    "funciton",
```
0x4007: This is obviously a typo of "function" you should fix all of the typos in the code instead of adding invalid words to this dictionary.

packages/contracts/src/dollar/ERC1155SetUri/ERC1155SetUri.sol
```diff
@@ -59,7 +66,13 @@ contract ERC1155SetUri is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
      * actual token type ID.
      */
-    function uri(uint256) public view virtual override returns (string memory) {
+    function uri(uint256)
+        public
+        view
+        virtual
+        override
+        returns (string memory)
+    {
```
0x4007: All of the formatting changes are unexpected. What formatter are you using here and how exactly are you invoking it?

cspell.json
```diff
@@ -36,7 +36,26 @@
...
+    "delegatecall",
+    "diamod",
+    "Ubistick",
```
0x4007: This should be UbiquiStick. Perhaps add "Ubiqui" as a word since Stick already is.

0x4007 (changes requested):
Fix the formatting before requesting another review.

0x4007:
@Draeieg now is the time to provide your math formula.

Draeieg:
> @Draeieg now is the time to provide your math formula.

I need the specifics mentioned in another ubq bonding curve issue, otherwise I will provide a formula with random undefined variables lol

1) how much of the current/projected ubq supply we want this to burn
2) are we still going for the 1024 target or should we go for more 
3) what's the maximum price at the end of the curve tapering 10 eth, 100 eth?

cspell.json
FibrinLab: I discovered most of the formatting and linting errors are not from the smart contracts I added to the repo. I have modified the ignore list further.

packages/contracts/src/dollar/ERC1155SetUri/ERC1155SetUri.sol
FibrinLab: Prettier. I started getting formatting errors when i tried to make a commit. I ran the Prettier format command as specified in the package.json file.

cspell.json
FibrinLab: Modified

FibrinLab:
> > @Draeieg now is the time to provide your math formula.
> 
> 
> 
> I need the specifics mentioned in another ubq bonding curve issue, otherwise I will provide a formula with random undefined variables lol
> 
> 
> 
> 1) how much of the current/projected ubq supply we want this to burn
> 
> 2) are we still going for the 1024 target or should we go for more 
> 
> 3) what's the maximum price at the end of the curve tapering 10 eth, 100 eth?

I guess this means I would implement a "Max Balance in the bonding curve vault contract"

Draeieg:
> I guess this means I would implement a "Max Balance in the bonding curve vault contract"

Sort of,  the curve I modeled is based on a logistic/sigmoid curve, which means the equation will naturally end up tapering to max value 

If you wish to implement that taper by directly doing the equation as I modeled them or want to just use a max value is up to you as long as there aren't any inherent problems with the method

FibrinLab:
> > I guess this means I would implement a "Max Balance in the bonding curve vault contract"
> 
> Sort of,  the curve I modeled is based on a logistic/sigmoid curve, which means the equation will naturally end up tapering to max value 
> 
> If you wish to implement that taper by directly doing the equation as I modeled them or want to just use a max value is up to you as long as there aren't any inherent problems with the method 
> 
> 

Would love to have a look

0x4007:
> > @Draeieg now is the time to provide your math formula.
> 
> I need the specifics mentioned in another ubq bonding curve issue, otherwise I will provide a formula with random undefined variables lol
> 
> 1. how much of the current/projected ubq supply we want this to burn
> 2. are we still going for the 1024 target or should we go for more
> 3. what's the maximum price at the end of the curve tapering 10 eth, 100 eth?

You should just explain the math here with your best assumptions since its easy to modify individual numbers later in the code. Subject to change, but 

1. if we can figure a way to (ideally in perpetuity) burn at a faster rate than we are emitting (1.5 per block) we're ultrasound money! Perhaps thats a bit too aspirational though. That could be a very interesting pricing algorithm, where it costs an increasing percentage of the entire UBQ supply with every subsequent mint, but it also feels a little strange to me how directly associated the UbiquiStick and UBQ supply would be. 
2. 1024 is fine.
3. 128 ETH seems like a fine number lol staying true to the number pattern.

FibrinLab:
> > > @Draeieg now is the time to provide your math formula.
> 
> > 
> 
> > I need the specifics mentioned in another ubq bonding curve issue, otherwise I will provide a formula with random undefined variables lol
> 
> > 
> 
> > 1. how much of the current/projected ubq supply we want this to burn
> 
> > 2. are we still going for the 1024 target or should we go for more
> 
> > 3. what's the maximum price at the end of the curve tapering 10 eth, 100 eth?
> 
> 
> 
> You should just explain the math here with your best assumptions since its easy to modify individual numbers later in the code. Subject to change, but 
> 
> 
> 
> 1. if we can figure a way to (ideally in perpetuity) burn at a faster rate than we are emitting (1.5 per block) we're ultrasound money! Perhaps thats a bit too aspirational though. That could be a very interesting pricing algorithm, where it costs an increasing percentage of the entire UBQ supply with every subsequent mint, but it also feels a little strange to me how directly associated the UbiquiStick and UBQ supply would be. 
> 
> 2. 1024 is fine.
> 
> 3. 128 ETH seems like a fine number lol staying true to the number pattern.

I will implement once I get the formulae.

Draeieg:
@FibrinLab 

### The Formula
we will be working with a standard logistic function where the formula is:
$\huge{\frac E {1+e^{-t(x-x_0)}}}$

### Variables/Functions explained

"E" is the maximum price in ETH the UbiquiStick will have this will be equal to $128$ ETH, so you can think of the formula as

>$\huge{\frac {128} {1+e^{-t(x-x_0)}}}$

the "t" function is what defines the rate at which the curve grows, for now this will be equal to $0.0015$ however this need to be upgradable to a fully fledged function as we're planning more esoteric methods of ramp up

$x_0$ is the Inflection point of the curve. In the same way as "t" this has to be upgradable from a flat number to a function, for now it equals 1800

$e$ is Euler's number which equals to $2.718281828$

$x$ is the number of the UbiquiStick to be minted at that point

> For example,  if were minting the 8th ubiquistick this will be equal to 
> 
> $\Large{\frac {128} {1+2.718281828^{-0.0015(8-1800)}}=8.1517021518}$
> 
> This means the 8th UbiquiStick should cost $8.1517021518$ ETH in UBQ to mint


you can find an example of how the curve works on a Graphing calculator to play around [Here](https://www.desmos.com/calculator/4xecjqccsp)

FibrinLab:
Thanks for providing this. I will work on implementing this with static values. Quick question, why an inflection point of 0.0015?

Draeieg:
> Thanks for providing this. I will work on implementing this with static values. Quick question, why an inflection point of 0.0015?

the Inflection point is $x_0$, which is 1800
$0.0015$ is the growth rate, this value was chosen because we needed the curve to grow slow but not too slow, this will make more sense if you play around the desmos simulation I made for you

also feel free to change those variable names in code to whatever you feel makes sense since its easier to write the math using those names but in code its not the same story

0x4007:
@0xcodercrane please take a look

0x4007:
@zgorizzo69 I understand that you were interested to implement the bonding curve but your attention is better spent on the other smart contracts. It would be lovely if you can add your opinion on here as I imagine that you have quite a bit of contextual knowledge on this matter.

cuddle-dev (changes requested):
I don't have much clear understanding on math so I wanted to go through everything with unit tests but its currently missing. Would you like to add unit tests to coverage all the functions and new files @FibrinLab ?

FibrinLab:
> I don't have much clear understanding on math so I wanted to go through everything with unit tests but its currently missing. Would you like to add unit tests to coverage all the functions and new files @FibrinLab ? 

I understand the basic math concepts, I will also try to add unit test. Let me work on a draft

FibrinLab:
@pavlovcik Please have a look at the smart contract.

0x4007:
Hey @FibrinLab unfortunately I'm not a Solidity developer so my review won't be very insightful. I hope that @zgorizzo69 can take a look because I know he has experience with bonding curves. I'll speak with him to see if I can get his input here soon. Thanks!

zgorizzo69:
> I temporarily had to create a prettier ignore file, I think there is a bug somewhere. I also added some more ignore words to the cspells ignore list. The smart contract is a fork with some new functions implements. Reviews appreciated.

fork from bancor from what I see

FibrinLab:
> > I temporarily had to create a prettier ignore file, I think there is a bug somewhere. I also added some more ignore words to the cspells ignore list. The smart contract is a fork with some new functions implements. Reviews appreciated.
> 
> 
> 
> fork from bancor from what I see

I am not done with the tests yet. I just want to know I am doing the right thing and some guidance

zgorizzo69:
> @FibrinLab
> ### The Formula
> 
> we will be working with a standard logistic function where the formula is: E1+e−t(x−x0)
> ### Variables/Functions explained
> 
> "E" is the maximum price in ETH the UbiquiStick will have this will be equal to 128 ETH, so you can think of the formula as
> 
> > 1281+e−t(x−x0)
> 
> the "t" function is what defines the rate at which the curve grows, for now this will be equal to 0.0015 however this need to be upgradable to a fully fledged function as we're planning more esoteric methods of ramp up
> 
> x0 is the Inflection point of the curve. In the same way as "t" this has to be upgradable from a flat number to a function, for now it equals 1800
> 
> e is Euler's number which equals to 2.718281828
> 
> x is the number of the UbiquiStick to be minted at that point
> 
> > For example,  if were minting the 8th ubiquistick this will be equal to
> > 1281+2.718281828−0.0015(8−1800)=8.1517021518
> > This means the 8th UbiquiStick should cost 8.1517021518 ETH in UBQ to mint
> 
> you can find an example of how the curve works on a Graphing calculator to play around [Here](https://www.desmos.com/calculator/4xecjqccsp)

@Draeieg 
Is it possible to make your formula bancor compliant ? 
indeed the bancor formula is l[ike this ](https://docs.google.com/spreadsheets/d/1szn6JuoaSwQevmfzvypt-psVTD2_snX62_12_CztuiQ/edit?usp=sharing)
by tweaking the connector weight we can have different kind of curve (i.e linear / exponential / log) 
this way we can really "fork" bancor and make sure the code is safe  

Here are my thoughts : 

- The formula depends on the number of ubiquiStick (NFT that contains a certain amount of UBQ)
so what happen if I don't have enough token to buy an NFT ? @pavlovcik you confirm that I should be able to deposit any amount right ?
- We would need other functions like If I deposit `x` amount of ETH (I understood that it could be any token but for simplicity we take its equivalent in ETH) I will get `y` amount of UBQ in my ubiquiStick 
- @FibrinLab we are using abdkMath everywhere else would be nice to use it here also instead of another implementation
for the power you can use   b ^ n == 2^(n*log²(b)) indeed 2^(nlog²(b)) = (2^(log²(b)))^n = b^n

FibrinLab:
> > @FibrinLab
> > ### The Formula
> > we will be working with a standard logistic function where the formula is: E1+e−t(x−x0)
> > ### Variables/Functions explained
> > "E" is the maximum price in ETH the UbiquiStick will have this will be equal to 128 ETH, so you can think of the formula as
> > > 1281+e−t(x−x0)
> > 
> > 
> > the "t" function is what defines the rate at which the curve grows, for now this will be equal to 0.0015 however this need to be upgradable to a fully fledged function as we're planning more esoteric methods of ramp up
> > x0 is the Inflection point of the curve. In the same way as "t" this has to be upgradable from a flat number to a function, for now it equals 1800
> > e is Euler's number which equals to 2.718281828
> > x is the number of the UbiquiStick to be minted at that point
> > > For example,  if were minting the 8th ubiquistick this will be equal to
> > > 1281+2.718281828−0.0015(8−1800)=8.1517021518
> > > This means the 8th UbiquiStick should cost 8.1517021518 ETH in UBQ to mint
> > 
> > 
> > you can find an example of how the curve works on a Graphing calculator to play around [Here](https://www.desmos.com/calculator/4xecjqccsp)
> 
> @Draeieg Is it possible to make your formula bancor compliant ? indeed the bancor formula is l[ike this ](https://docs.google.com/spreadsheets/d/1szn6JuoaSwQevmfzvypt-psVTD2_snX62_12_CztuiQ/edit?usp=sharing) by tweaking the connector weight we can have different kind of curve (i.e linear / exponential / log) this way we can really "fork" bancor and make sure the code is safe
> 
> Here are my thoughts :
> 
> * The formula depends on the number of ubiquiStick (NFT that contains a certain amount of UBQ)
>   so what happen if I don't have enough token to buy an NFT ? @pavlovcik you confirm that I should be able to deposit any amount right ?
> * We would need other functions like If I deposit `x` amount of ETH (I understood that it could be any token but for simplicity we take its equivalent in ETH) I will get `y` amount of UBQ in my ubiquiStick
> * @FibrinLab we are using abdkMath everywhere else would be nice to use it here also instead of another implementation
>   for the power you can use   b ^ n == 2^(n*log²(b)) indeed 2^(nlog²(b)) = (2^(log²(b)))^n = b^n

@Draeieg @zgorizzo69 I will wait for the updated formula then refactor.

0x4007:
> * you confirm that I should be able to deposit any amount right ?

Should be fine unless @Draeieg says otherwise.

> @Draeieg @zgorizzo69 I will wait for the updated formula then refactor.

Looks like @zgorizzo69 already provided the answer unless I'm misunderstanding his comment.

zgorizzo69:
> 

in the current design we have to spend a certain amount of ETH to buy a stick my assumption was that no matter how much you spend you should get a stick is that right @pavlovcik ?

FibrinLab:
> > 
> 
> in the current design we have to spend a certain amount of ETH to buy a stick my assumption was that no matter how much you spend you should get a stick is that right @pavlovcik ?

@pavlovcik should I fork the bancor formula directly or wait for the updated one?

0x4007:
> in the current design we have to spend a certain amount of ETH to buy a stick my assumption was that no matter how much you spend you should get a stick is that right @pavlovcik ?

I think its fine.

> @pavlovcik should I fork the bancor formula directly or wait for the updated one?

@Draeieg you should make the formula Bancor compliant to ease implementation.

Draeieg:
1) Making this bancor compliant was more annoying than expected because it wasn't designed for our use case and as a results lacks features and the easy route is going to rekt some of the usability that bancor provides 

for now we're going to take the easy route, all data for the updated formula is in this [Sheet](https://docs.google.com/spreadsheets/d/1LLM0tIEF5NMyO3D10Cyecwbo3xzKcgeu5csJ_hWfXMw/edit?usp=sharing), the supply that will be taken into account will be the supply of UbiquiStick NOT the supply of ubq 

however keep in mind if possible that, in an ideal world we would upgrade this to a Bonding ***Surface*** instead of a Bonding ***Curve*** for all of the usability we have planned
2) No, each UbiquiStick will COST the amount defined in the formula, only by paying that you get the stick. 
	
>putting any amount would be like going to buy a car and pay whatever you want lmao. 
	only way to be able to deposit any amount would be to fractionalize the UbiquiStick, seems like a cool concept taking into consideration our whole cyberpunk style but complicated to implement IMO

this is more UI than curve itself but it should calculate the deposit amount needed for 1 UbiquiStick returned instead o tokens returned as the token returned should be a constant 1 and the variable would actually be how much money is needed

0x4007:
> Bonding _**Surface**_

This is a new term for me. Can you elaborate?

> updated formula is in this [Sheet](https://docs.google.com/spreadsheets/d/1LLM0tIEF5NMyO3D10Cyecwbo3xzKcgeu5csJ_hWfXMw/edit?usp=sharing)

For archival purposes I recommend embedding all external files here. In this case, a screenshot can suffice.

image

FibrinLab:
> 1. Making this bancor compliant was more annoying than expected because it wasn't designed for our use case and as a results lacks features and the easy route is going to rekt some of the usability that bancor provides
> 
> for now we're going to take the easy route, all data for the updated formula is in this [Sheet](https://docs.google.com/spreadsheets/d/1LLM0tIEF5NMyO3D10Cyecwbo3xzKcgeu5csJ_hWfXMw/edit?usp=sharing), the supply that will be taken into account will be the supply of UbiquiStick NOT the supply of ubq
> 
> however keep in mind if possible that, in an ideal world we would upgrade this to a Bonding _**Surface**_ instead of a Bonding _**Curve**_ for all of the usability we have planned 2) No, each UbiquiStick will COST the amount defined in the formula, only by paying that you get the stick.
> 
> > putting any amount would be like going to buy a car and pay whatever you want lmao.
> > only way to be able to deposit any amount would be to fractionalize the UbiquiStick, seems like a cool concept taking into consideration our whole cyberpunk style but complicated to implement IMO
> 
> this is more UI than curve itself but it should calculate the deposit amount needed for 1 UbiquiStick returned instead o tokens returned as the token returned should be a constant 1 and the variable would actually be how much money is needed

Working on this

Draeieg:
> > Bonding _**Surface**_
> 
> This is a new term for me. Can you elaborate?
> 
> > updated formula is in this [Sheet](https://docs.google.com/spreadsheets/d/1LLM0tIEF5NMyO3D10Cyecwbo3xzKcgeu5csJ_hWfXMw/edit?usp=sharing)
> 
> For archival purposes I recommend embedding all external files here. In this case, a screenshot can suffice.
> 
> image

A curve is a 2d graph with 2 variables, X and Y. In this case supply of UbiquiSticks and as a result price
A surface is a 3d graph with 3 variables, X, Y and Z. supply of ubiquisticks supply of UBQ and price as aresult

the surface would be the example mentioned a few weeks ago where I proposed to move the inflection point of the logistic curve depending on the supply of ubq to incentivize/disincentivize buying the ubiquistick depending on how much UBQ we would like to burn

FibrinLab:
> Sheet

Can I have access to this sheet. Thanks

FibrinLab:
@pavlovcik @zgorizzo69 Updated the PR to become more bancor compliant. Waiting for reviews. The connector weights and baseY are part of the constructor rather that being hardcoded values.

FibrinLab:
@pavlovcik Still waiting for a review. Will this require initial test for now?

0x4007:
Hey @zgorizzo69 do you have any remarks on this? This is more up your alley.

0x4007:
> Can I have access to this sheet. Thanks 

I gave access.

packages/contracts/src/dollar/BondingCurve.sol
```diff
@@ -12,7 +13,7 @@ import "./core/UbiquityDollarManager.sol";
  * Inspired from Bancor protocol and simondlr
  * https://github.com/bancorprotocol/contracts
  */
-contract BondingCurve is BancorFormula {
+contract BondingCurve is BancorFormula, Pausable {
     uint256 constant ACCURACY = 10 ** 18;
```
zgorizzo69: you use 10**18 but we mostly use exponentiation with the `e` expression e.g 1e18
While both expressions result in the same values, I recommend always using the same `e` to avoid potential confusion. Also  a constant expression in a variable will compute the expression everytime the variable is called. It's not the result of the expression that is stored, but the expression itself. So 1e18 will save gas

zgorizzo69:
I don't see any withdraw function so it means that we will never be able to burn a stick to get the collateral out right @Draeieg  ?

Draeieg:
> I don't see any withdraw function so it means that we will never be able to burn a stick to get the collateral out right @Draeieg  ?
> 

Yes that's correct. If they want to ape out gotta find someone else on opensea

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.3;
```
zgorizzo69: it feels empty ...
can we add some tests to make sure that for different collateral deposit we have accurate tokens returned (using  different connecteor weight and baseY)

zgorizzo69 (changes requested):

0x4007:
Following up on this @zgorizzo69 @FibrinLab

FibrinLab:
> Following up on this @zgorizzo69 @FibrinLab 

Working on an update. I need a few  hours.

zgorizzo69:
> > Following up on this @zgorizzo69 @FibrinLab
> 
> Working on an update. I need a few hours.

ok just ping me then

FibrinLab:
@pavlovcik @zgorizzo69 
Thanks for your patience in this PR. I made the following updates.
1. Added tests for the bonding curve contract covering all basic functions.
2. Moved bonding curve parameters to functions so they can my modified with adequate permissions.
3. Implemented access controls for functions.
4. Added a "Withdraw" function for deposited tokens can still be recovered.

Sorry for the delay. Had to research a little. 🧐

zgorizzo69:
@FibrinLab bondign test is failing

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,123 @@
...
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "../helpers/LiveTestHelper.sol";
+
+contract ZeroState is LiveTestHelper {
```
zgorizzo69: :+1:

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,123 @@
...
+            _baseY
+        ); 
+
+        bondingCurve.setCollateralToken(
+            address(dollarToken)
+        );
+
+        uint256 tokenIds = 0;
+        uint256 poolBalance = bondingCurve.poolBalance();
+
```
zgorizzo69: we can stop the prank here as deposit is not restricted to admin

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,123 @@
...
+        bondingCurve.deposit(
+            collateralDeposited, 
+            secondAccount
+        );
+
```
zgorizzo69: same here we can stop prank

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,123 @@
...
+        uint256 tokenIds = 0;
+        uint256 poolBalance = bondingCurve.poolBalance();
+
+        bondingCurve.deposit(
+            collateralDeposited, 
+            secondAccount
+        );
+        vm.stopPrank();
+
```
zgorizzo69: we should assert that we bow have ubiquistick and the right amount of it ..

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,123 @@
...
+        bondingCurve.deposit(
+            collateralDeposited, 
+            secondAccount
+        );
+
+        bondingCurve.withdraw(
+            thirdAccount,
+            collateralDeposited
+        );
+        vm.stopPrank();
```
zgorizzo69: are we forced to withdraw all ? or can we only remove 1 ubiquistick at a time ?
we should assert that now we don't have ubiquistick and or that we are left with the right amountt ..

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,123 @@
...
+
+    uint256 collateralDeposited;
+
+    function testDeposit() public {
+        vm.expectEmit(true, false, false, true);
+        emit Deposit(secondAccount, collateralDeposited);
+
+        vm.startPrank(admin);
+        bondingCurve.setParams(
+            _connectorWeight,
```
zgorizzo69: would be very neat to have a fuzzing test here with _connectorWeight and _baseY as input and we calculate based on this input that we receive the right amount of ubiquistick

packages/contracts/src/dollar/BondingCurve.sol
```diff
@@ -0,0 +1,164 @@
...
+
+    function pause() public virtual onlyPauser {
+        _pause();
+    }
+
+    function unpause() public virtual onlyPauser {
+        _unpause();
+    }
+
+    function withdraw(address _to, uint256 _amount) external onlyBondingMinter {
```
zgorizzo69: so it is a withdraw all collateral function. @pavlovcik don't you think we should restrict the withdraw to the DAO treasury (you can get the address from the manager)  ?

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: Fixed.

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: fixed

packages/contracts/src/dollar/BondingCurve.sol
FibrinLab: I am just gonna add a function to set the treasure address.

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: Added an assertion

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: done

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: Added assertion. Withdrawal amount can be specified

FibrinLab:
@zgorizzo69 @pavlovcik 
1. CI now passing.
2. All assertions updated as recommended
3. Withdrawal to treasury implemented

packages/contracts/src/dollar/BondingCurve.sol
0x4007: [Treasury](https://etherscan.io/address/fisc.ubq.eth) is a 4/4 multisig and is tedious to do operations with. The "admin" [wallet](https://etherscan.io/address/ubq.eth) is a MetaMask hot wallet which is far more convenient to do multi-transaction-operations with.

Could be nice to "withdraw to treasury" but I would consider the capability to withdraw to the "admin" wallet is more essential.

FibrinLab: @pavlovcik should i revert back to just a simple function that sets "admin" wallet?

packages/contracts/src/dollar/BondingCurve.sol
```diff
@@ -0,0 +1,183 @@
...
+
+    constructor(
+        address _manager,
+        address _token
+    ) {
+        require(_token != address(0), "NFT address empty");
+        tokenAddr = _token;
+        token = IUbiquiStick(_token);
+
+        manager = UbiquityDollarManager(_manager);
```
zgorizzo69: here you should replace by IAccessControl cos we don't have it anymore

packages/contracts/src/dollar/BondingCurve.sol
zgorizzo69: like I said no need to set treasury cos it is inside the Manager already you can do like this :
IUbiquityDollarManager(_manager).treasuryAddress()

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,135 @@
...
+    function testCannotDeployConnectorWeightZero() public {
+        vm.expectRevert();
+        BondingCurve broken = new BondingCurve(
+            address(manager),
+            address(0)
+        );
+    }
+
+
+    uint256 connWeight = bound(uint256(connectorWeight), 1, 1000000);
```
zgorizzo69: to make fuzzing works it should be parameters of the functions like 
```

    function testSetParams(uint256 connectorWeight, uint256 baseY) public {
        uint256 connWeight = bound( connectorWeight , 1, 1000000);
        uint32 _connectorWeight = uint32(connWeight);
        uint256 _baseY = bound(baseY, 1, 1000000);
        vm.prank(admin);
        bondingCurve.setParams(
            _connectorWeight,
            _baseY
        );

        assertEq(bondingCurve.connectorWeight(), _connectorWeight);
        assertEq(bondingCurve.baseY(), _baseY);
    }

```

packages/contracts/test/dollar/BondingCurve.t.sol
```diff
@@ -0,0 +1,135 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.3;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "../helpers/LiveTestHelper.sol";
+
+contract ZeroState is LiveTestHelper {
```
zgorizzo69: why is it a liveTest ? can't it be a localTest ?

zgorizzo69 (changes requested):

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: I noticed most of the contracts make use of the live tests so i just followed the pattern. Any recommendations? @zgorizzo69

zgorizzo69: you mean most of them use LocalTester right ? 
I would prefer to use LocalTester especially when we don't have a dependency on the live mainnet

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: Done

packages/contracts/src/dollar/BondingCurve.sol
FibrinLab: Is this a new update? AccessControl is already declared in the UbiquityDollarManager

zgorizzo69: yep diamond will replace some of the current contracts including manager

FibrinLab: @zgorizzo69 Since we are moving to diamond, can I just use the ManagerFacet from the Diamond all together? Or rather move the BondingCure to Diamond. What do you suggest?

packages/contracts/test/dollar/BondingCurve.t.sol
FibrinLab: Done. Will be available in next commit

packages/contracts/src/dollar/BondingCurve.sol
zgorizzo69: BondingCurve as a facet makes sense !

FibrinLab: I will have it ready by tomorrow

packages/contracts/test/dollar/BondingCurve.t.sol
zgorizzo69: where is it done ? Did you  push already ?

FibrinLab:
@zgorizzo69 @pavlovcik I have moved BondingCurve and Bancor formula to diamond. Had to do a bit of research. Still adapting 1 more test. Review will be appreciated .
Thanks for your patience.

FibrinLab:
@zgorizzo69 Moved bonding to diamond with tests. Thanks

packages/contracts/src/dollar/core/BancorFormula.sol
```diff
@@ -0,0 +1,105 @@
...
+     * @dev _supply * ((1 + _tokensDeposited / _connectorBalance) ^ (_connectorWeight / 1000000) - 1)
+     *
+     * @param _tokensDeposited   amount of collateral tokens to deposit
+     * @param _connectorWeight   connector weight, represented in ppm, 1 - 1,000,000
+     * @param _supply          current Token supply
+     * @param _connectorBalance   total connector balance
+     * 
+     * @return amount of Tokens minted
+     */
+    function _purchaseTargetAmount(
```
zgorizzo69: I think we can remove this contract as it is in the facte now

packages/contracts/src/dollar/facets/BancorFormulaFacet.sol
```diff
@@ -0,0 +1,69 @@
...
+import {Modifiers} from "../libraries/LibAppStorage.sol";
+
+/**
+ * @title Bancor formula
+ * @dev Modified from the original
+ * @notice forked from Bancor:
+ * https://github.com/bancorprotocol/contracts
+ *
+ */
+contract BancorFormulaFacet is Modifiers {
```
zgorizzo69: same contract name inside facets/bancorformulaFacet

zgorizzo69 (changes requested):

packages/contracts/src/dollar/core/BancorFormula.sol
FibrinLab: alright

packages/contracts/src/dollar/facets/BancorFormulaFacet.sol
FibrinLab: Do you want the name modified?

packages/contracts/src/dollar/core/BancorPower.sol
```diff
@@ -0,0 +1,380 @@
...
+
+/**
+ * bancor formula by bancor
+ * https://github.com/bancorprotocol/contracts
+ * Modified from the original
+ * Split Power.sol out from BancorFormula.sol
+ * Licensed to the Apache Software Foundation (ASF) under one or more contributor license agreements;
+ * and to You under the Apache License, Version 2.0. "
+ */
+contract BancorPower {
```
zgorizzo69: I don't see where this contract is used so we can remove it

packages/contracts/src/dollar/libraries/LibBancorFormula.sol
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+
+import {BancorPower} from "../core/BancorPower.sol";
```
zgorizzo69: not used

packages/contracts/src/dollar/facets/BancorFormulaFacet.sol
```diff
@@ -0,0 +1,69 @@
...
+     * @dev _supply * ((1 + _tokensDeposited / _connectorBalance) ^ (_connectorWeight / 1000000) - 1)
+     *
+     * @param _tokensDeposited   amount of collateral tokens to deposit
+     * @param _connectorWeight   connector weight, represented in ppm, 1 - 1,000,000
+     * @param _supply          current Token supply
+     * @param _connectorBalance   total connector balance
+     * 
+     * @return amount of Tokens minted
+     */
+    function _purchaseTargetAmount(
```
zgorizzo69: we can remove the underscore in the name as it is an external func

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
```diff
@@ -0,0 +1,114 @@
...
+            baseY
+        ); 
+
+        IBondingCurveFacet.deposit(
+            collateralDeposited, 
+            secondAccount
+        );
+        vm.stopPrank();
+
+        assertEq(IBondingCurveFacet.poolBalance(), collateralDeposited);
```
zgorizzo69: we miss some tests in deposit and withdraw :
* check that the bFT has been minted correctly
* check that the amount of token returned is correct for amount AND amountFromzero

packages/contracts/src/dollar/libraries/LibBancorFormula.sol
```diff
@@ -0,0 +1,112 @@
...
+     * @dev _supply * ((1 + _tokensDeposited / _connectorBalance) ^ (_connectorWeight / 1000000) - 1)
+     *
+     * @param _tokensDeposited   amount of collateral tokens to deposit
+     * @param _connectorWeight   connector weight, represented in ppm, 1 - 1,000,000
+     * @param _supply          current Token supply
+     * @param _connectorBalance   total connector balance
+     * 
+     * @return amount of Tokens minted
+     */
+    function _purchaseTargetAmount(
```
zgorizzo69: we can merge this lib with the bonding curve lib

packages/contracts/src/dollar/facets/BancorFormulaFacet.sol
zgorizzo69: actually we don't need that facet has no contract will need an external call to these funcs 
so better merge the libbancorformula with libbonding curve

zgorizzo69 (changes requested):

packages/contracts/src/dollar/facets/BancorFormulaFacet.sol
FibrinLab: Done

packages/contracts/src/dollar/libraries/LibBancorFormula.sol
FibrinLab: Done

packages/contracts/src/dollar/libraries/LibBancorFormula.sol
FibrinLab: Done

packages/contracts/src/dollar/core/BancorPower.sol
FibrinLab: Removed

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
FibrinLab: Resolved

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
```diff
@@ -72,7 +74,15 @@ contract ZeroStateStakingTest is BondingCurveFacetTest {
         );
         vm.stopPrank();
 
+        uint256 tokReturned = IBondingCurveFacet.purchaseTargetAmountFromZero(
```
zgorizzo69, line 101: these tests are fine but we also need
- another test that will use purchaseTargetAmount
- other tests to make sure  purchaseTargetAmountFromZero and purchaseTargetAmount returns correct calculations

zgorizzo69 (changes requested):

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
FibrinLab: I made some corrections

FibrinLab:
@zgorizzo69 I appreciate the review. I added some more tests. Cool learning experience.

packages/contracts/src/dollar/facets/BondingCurveFacet.sol
```diff
@@ -46,7 +46,7 @@ contract BondingCurveFacet is Modifiers, IBondingCurve {
         return LibBondingCurve.getShare(_recipient);
     }
 
-    function withdraw(uint256 _amount) external whenNotPaused {
+    function withdraw(uint256 _amount) external onlyAdmin whenNotPaused {
```
zgorizzo69, line 49: can you clarify why it is admin only ?

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
```diff
@@ -97,9 +97,10 @@ library LibBondingCurve {
             _collateralDeposited
         );
 
-        ss.poolBalance = ss.poolBalance + _collateralDeposited;
+        ss.poolBalance += _collateralDeposited;
```
zgorizzo69: btw it uses less gas NOT to use +=

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
```diff
@@ -53,16 +56,31 @@ contract ZeroStateBonding is BondingCurveFacetTest {
         assertEq(baseY, IBondingCurveFacet.baseY());
     }
 
+    function testSetParamsShouldRevertNotAdmin() public {
```
zgorizzo69, line 62: :+1:

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
```diff
@@ -125,28 +145,66 @@ contract ZeroStateBonding is BondingCurveFacetTest {
...
+        );
+    }
+
+    function testPurchaseTargetAmount() public {
+        // Calculate expected result
+        uint256 tokensDeposited;
+        uint connWeight;
+        uint256 tokenIds;
+        uint32 connectorWeight = uint32(bound(connWeight, 1, MAX_WEIGHT));
+        uint256 bal;
```
zgorizzo69: can provide this to input parameter so that this test is a fuzzing test

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
zgorizzo69: Please point me where the test calling purchaseTargetAmount is ?
and here we only assess that tokreturned is what getshare returns but we need to assess that it is equal to an expected value that we need to calculate here with  _supply * ((1 + _tokensDeposited / _connectorBalance) ^ (_connectorWeight / 1000000) - 1)

zgorizzo69 (changes requested):

packages/contracts/src/dollar/facets/BondingCurveFacet.sol
FibrinLab: I assumed only the admin would have the right and permission to withdraw to the treasury

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
FibrinLab: Fixed

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
FibrinLab: Got IT

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
FibrinLab: It has a seperate function. Currently line 189. Thanks for the review. Updated.

FibrinLab:
@zgorizzo69 Review reminder

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
```diff
@@ -102,8 +105,14 @@ contract ZeroStateBonding is BondingCurveFacetTest {
...
         );
 
+        // Logic Test
+        uint256 baseN = collateralDeposited.add(baseY);
+        uint256 power = (baseN.mul(10 ** 18)).div(baseY);
+        uint256 result = ACCURACY.mul(SafeMath.sub((power**(connectorWeight)), 10 ** 18)).div(10 ** 18);
+
         assertEq(collateralDeposited, IBondingCurveFacet.poolBalance());
         assertEq(collateralDeposited, finBal - initBal);
+        assertEq(tokReturned, result);
```
zgorizzo69, line 115: :+1:

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
```diff
@@ -177,13 +186,11 @@ contract ZeroStateBonding is BondingCurveFacetTest {
         );
     }
 
-    function testPurchaseTargetAmount() public {
+    function testPurchaseTargetAmount(uint32 connWeight, uint256 bal) public {
```
zgorizzo69, line 189: :+1:

zgorizzo69 (approved):

packages/contracts/src/dollar/interfaces/IBancorFormula.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.3;
```
0x4007, line 2: Shouldn't all of these be updated to the latest Solidity version?

packages/contracts/src/dollar/libraries/Constants.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.16;
 
+import "abdk/ABDKMathQuad.sol";
```
0x4007, line 4: I was under the impression that the later versions of Solidity (including the one in this file pragma) makes ABDK library obsolete?

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
```diff
@@ -0,0 +1,235 @@
...
+        uint256 baseY;
+        uint256 poolBalance;
+        uint256 tokenIds;
+        mapping (address => uint256) share;
+    }
+
+    function bondingCurveStorage() internal pure returns (BondingCurveData storage l) {
+        bytes32 slot = BONDING_CONTROL_STORAGE_SLOT;
+        assembly {
+            l.slot := slot
```
0x4007, line 37: What's up with the assembly? I presume that this operation is not possible to do in Solidity for some reason?

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    function getShare(address _recipient) internal returns (uint256) {
+        BondingCurveData storage ss = bondingCurveStorage();
+        return ss.share[_recipient];
+    }
+
+    function toBytes(uint256 x) internal pure returns (bytes memory b) {
+        b = new bytes(32);
+        assembly { mstore(add(b, 32), x) }
+    }
```
0x4007, line 126: Also curious about this function. Its unreadable to me.

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    function withdraw(uint256 _amount) internal {
+        BondingCurveData storage ss = bondingCurveStorage();
```
0x4007, line 129: Why is this variable named `ss` is this forked from our old staking shares code?

packages/contracts/src/dollar/mocks/MockUbiquistick.sol
```diff
@@ -0,0 +1,15 @@
+// contracts/GLDToken.sol
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.16;
+
+import "@openzeppelin/contracts/token/ERC1155/ERC1155.sol";
+
+contract MockUbiquistick is ERC1155 {
+    // cspell: disable-next-line
+    constructor() ERC1155("ubiquiStick") {
```
0x4007:
```suggestion
    constructor() ERC1155("UbiquiStick") {
```

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
FibrinLab: No it's not forked. Just following the diamond facet schema.

packages/contracts/test/diamond/DiamondTest.t.sol
```diff
@@ -12,7 +12,7 @@ contract TestDiamond is DiamondSetup {
     }
 
     function testHasMultipleFacets() public {
-        assertEq(facetAddressList.length, 15);
+        assertEq(facetAddressList.length, 16);
     }
```
0x4007, line 16: I feel like this test will be tricky if we are merging in new branches/features constantly. Perhaps it should just log the amount of facets? @zgorizzo69

packages/contracts/test/diamond/DiamondTestSetup.sol
```diff
@@ -436,7 +483,7 @@ abstract contract DiamondSetup is DiamondTestHelper {
             )
         });
 
-        FacetCut[] memory cuts = new FacetCut[](15);
+        FacetCut[] memory cuts = new FacetCut[](16);
```
0x4007, line 486: Also here, related to my previous comment above.

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
```diff
@@ -0,0 +1,217 @@
...
+        ); 
+
+        IBondingCurveFacet.deposit(
+            collateralDeposited, 
+            secondAccount
+        );
+
+        uint256 _amount = bound(baseY, 0, collateralDeposited);
+
+        vm.expectEmit(true, false, false, true);
```
0x4007, line 143: What does `true, false, false, true` represent?

packages/contracts/test/diamond/DiamondTestSetup.sol
FibrinLab: Could be updated with new branches

0x4007 (commented):
I only focused on code style and asked a couple of questions because I'm not a Solidity developer.

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
0x4007: What does `ss` stand for?

FibrinLab:
> I only focused on code style and asked a couple of questions because I'm not a Solidity developer.

Nice. Will answer each individually.

packages/contracts/test/diamond/DiamondTestSetup.sol
0x4007: My fear is when two branches are both looking to merge in a new facet and both have incremented the facet count by one. The first one will pass the merge CI. The second one will fail. It is a simple problem but technically it would be a perpetual problem. That's why I'm bringing this to the team's attention to see if there's a better way to do this.

packages/contracts/src/dollar/interfaces/IBancorFormula.sol
FibrinLab: It uses the latest version available in the 0.8 range

packages/contracts/test/diamond/facets/BondingCurveFacet.t.sol
FibrinLab: This explains it nicely 

https://book.getfoundry.sh/cheatcodes/expect-emit

packages/contracts/src/dollar/libraries/Constants.sol
FibrinLab: @zgorizzo69 created that contract. Should be able to provide more info.

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
FibrinLab: It's a function that converts a uint256 to bytes32.

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
FibrinLab: It's actually possible. Low level call. It's part of the diamond facet architecture

packages/contracts/src/dollar/mocks/MockUbiquistick.sol
FibrinLab: Will make a commit for this. I don't want conventional commit to fail.

packages/contracts/test/diamond/DiamondTestSetup.sol
FibrinLab: @zgorizzo69

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
0x4007: Shouldn't it be implemented in Solidity if its possible to do in Solidity for readability purposes?

packages/contracts/src/dollar/mocks/MockUbiquistick.sol
0x4007: I'm pretty sure that @Steveantor updated our conventional commits to support suggestions from the GitHub reviews process!

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
FibrinLab: I don't think it's possible because it's allocating a memory slot

packages/contracts/src/dollar/mocks/MockUbiquistick.sol
FibrinLab: > I'm pretty sure that @Steveantor updated our conventional commits to support suggestions from the GitHub reviews process! 

Just accepted the commit.

0x4007:
I see it failed. It could be that your CI code is out of date on this branch. Here's proof that we updated it to support GitHub review suggestion commits: https://github.com/ubiquity/ubiquity-dollar/pull/623

packages/contracts/src/dollar/libraries/Constants.sol
zgorizzo69: Nop abdk is for floating point math which doesn't exist in solidity

FibrinLab:
> I see it failed. It could be that your CI code is out of date on this branch. Here's proof that we updated it to support GitHub review suggestion commits: #623

Its actually working fine

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
0x4007: @zgorizzo69 what do you think?

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
FibrinLab: I just randomly picked it following the patter of other facets. I could have equally called it "bs", "bonding storage"

packages/contracts/src/dollar/libraries/LibBondingCurve.sol
zgorizzo69: There is no "solidity" keyword to allocate memory slot only asm

Steveantor: > It's actually possible. Low level call. It's part of the diamond facet architecture

shouldn't we build on top of [solidstate](https://github.com/solidstate-network/solidstate-solidity)?

FibrinLab: @Steveantor solidstate uses the diamond pattern likewise this repo

Steveantor: > @Steveantor solidstate uses the diamond pattern likewise this repo

yes, we shouldn't try to reinvent the wheel if there's something better and well-recognized.
it could be built on top of solid-state, what's the reason for not doing so?

Steveantor: solid-state is like the openzeppelin for diamond

FibrinLab: > > @Steveantor solidstate uses the diamond pattern likewise this repo
> 
> 
> 
> yes, we shouldn't try to reinvent the wheel if there's something better and well-recognized.
> 
> it could be built on top of solid-state, what's the reason for not doing so?

I followed the pattern to add the bonding curve lib and facet. If we are switching to solid state it would have to be implemented for other facets and libs in the entire repo. Thoughts?

Steveantor: > > > @Steveantor solidstate uses the diamond pattern likewise this repo
> > 
> > 
> > yes, we shouldn't try to reinvent the wheel if there's something better and well-recognized.
> > it could be built on top of solid-state, what's the reason for not doing so?
> 
> I followed the pattern to add the bonding curve lib and facet. If we are switching to solid state it would have to be implemented for other facets and libs in the entire repo. Thoughts?

I see your reasoning @zgorizzo69  may have more input

0x4007: I don't like the idea of introducing nascent frameworks unless it's absolutely necessary. There are always security risks with introducing new technologies, but if they are widely used (like Open Zeppelin libraries) it is more likely that the vulnerabilities would have already been exploited and patched.

To be honest I haven't heard of it before this thread, but I follow a lot of top level Solidity developers on Twitter and usually the highest quality technologies land on my radar there first.

FibrinLab: @pavlovcik if we are adding a diamond framework we would have to probably make an issue and do it for all facets

FibrinLab:
@pavlovcik Any other modifications from this PR?

0x4007 (approved):

0x4007:
@0xcodercrane

cuddle-dev (approved):

PR #455, opened 2023-01-05
fix: lint-staged for solidity

### LINKED ISSUE #447: Solidity Prettier Misconfiguration
image

Source: https://t.me/UbiquityDevPool/2303

> Lint-staged might be misconfigured,
I had use --no-verify a few times.

> We should ideally verify that it’s misconfigured and then fix it

> Although when I tried to make a commit today I didn’t encounter any issues again

> Perhaps it’s only for solidity files and not typescript files? I wouldn’t really know as I never touch the solidity code

> Yes. I agree. Prettier has an override for .sol files. I think that’s what is causing it

Steveantor:
- install `prettier-plugin-solidity` for prettier
- update `.prettierrc `
- update _lint-staged_ in `package.json`

Resolves #447

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -5,7 +5,9 @@
   "singleQuote": false,
   "printWidth": 160,
   "htmlWhitespaceSensitivity": "strict",
-  "plugins": [],
+  "plugins": [
+    "prettier-plugin-solidity"
+  ],
   "overrides": [
     {
       "files": "*.sol",
@@ -14,8 +16,7 @@
         "tabWidth": 4,
         "useTabs": false,
         "singleQuote": false,
-        "bracketSpacing": false,
-        "explicitTypes": "always"
+        "bracketSpacing": false
       }
     }
   ]
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "husky": "^8.0.2",
     "lint-staged": "^13.0.3",
     "nodemon": "^2.0.19",
+    "prettier-plugin-solidity": "^1.1.1",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -38,10 +39,14 @@
     "postinstall": "husky install"
   },
   "lint-staged": {
-    "*.{js,jsx,ts,tsx,json,sol}": [
+    "*.{js,jsx,ts,tsx,json}": [
       "prettier --write",
       "eslint --fix",
       "cspell --no-must-find-files"
+    ],
+    "*.sol": [
```
0x4007, line 47: Logically, this is the same. What's the difference?

.prettierrc
```diff
@@ -14,9 +16,8 @@
         "tabWidth": 4,
         "useTabs": false,
         "singleQuote": false,
-        "bracketSpacing": false,
-        "explicitTypes": "always"
```
0x4007, line 18: I guess this applies for everything but *.sol ?

Steveantor: It only applies to *.sol,
_explicitTypes_ is unsupported option for solidity linter.

Steveantor: > I guess this applies for everything but *.sol ?

It only applies to *.sol,
_explicitTypes_ is unsupported option for solidity linter. 
@pavlovcik

package.json
Steveantor: > Logically, this is the same. What's the difference?

not running `eslint --fix` with *.sol files.
@pavlovcik

0x4007 (approved):

PR #463, opened 2023-01-10
uCR-NFT ⮕ uCR Conversion

### LINKED ISSUE #400: `uCR-NFT` ⮕ `uCR` Conversion
Allow users to convert Ubiquity Credit NFTs to Ubiquity Credits (fungible).

Depends on #115 (we need to know the redemption rate of `1 uCR` for `? uAD` as it has less redemption power over time)

## Background

- We know that uCR-NFT is always worth 1 uAD
- We know how much uAD is 1 uCR worth at any given point in time
  - Depends on #115
- We know what duration of the uCR-NFT remains in percentage

## Math
 
 ```
(1 uCR-NFT) * (Remaining duration %) * (uCR/uAD conversion rate)
```

> Example:
> 
> 10 000 uCR-NFT with a total duration of 6 months has been around for 4 months, so they have 33% duration remaining.
> uCR/uAD redemption rate is 1.5
> So the user can get: 10 000 * 33% * 1.5 = 4 950 uCR (which is currently worth 3300 uAD)

_Originally posted by @hpiyankov in https://github.com/ubiquity/ubiquity-dollar/issues/105#issuecomment-1115346883_
      
## Note

In #112 it is emphasized where to implement this function

> Right now if you convert uCR-NFT to uCR with the CreditManager you get a 1-1 conversion, no > mater how close the uCR-NFT token is to expiring. While if expired does not allow you to > convert at all.
> 
> The relevant function is burnCreditNftsForCredits
> 
> What needs to happen
> Using the CreditClock we should consider the price of the uAD-uAD-equivalent of the uCR-NFT > amount (right now it converts 1-1) (this means dividing it by the CreditClock > creditRateAtBlockNumber, the result should be a higher amount of uCR (because it's devalued))
> At mint time the conversion should be 100%
> At expiration time the conversion should be 0%
> In between we lower it linearly using the expiration time

### PR DESCRIPTION
**Pull request by ethernomad**
Resolves #400

### CODE REVIEW

File: packages/contracts/src/dollar/core/CreditClock.sol [hunk 1]
```diff
@@ -0,0 +1,88 @@
...
+    // Manager contract.
+    UbiquityDollarManager private immutable manager;
+
+    // ABDKMathQuad with value of 1.
+    bytes16 private immutable one = uint256(1).fromUInt();
+
+    // The block height from where we start applying the rate.
+    uint256 public rateStartBlock;
+
+    // This is the exchange rate of uAR for the start block.
```
0x4007, line 20:
```suggestion
    // This is the exchange rate of Ubiquity Credits for the start block.
```

0x4007: We do not allow for the use of token symbol references in our code because we have the ability to update them, as well as to enhance clarity for new developers.

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -0,0 +1,88 @@
...
+    bytes16 public rateStartValue;
+
+    // Deprecation rate. How much uAD is deprecated on each block.
```
0x4007, line 23:
```suggestion
    // Deprecation rate. How much Ubiquity Dollars is deprecated on each block.
```

packages/contracts/src/dollar/core/CreditClock.sol
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.3;
+
+import "../libs/ABDKMathQuad.sol";
```
0x4007, line 4: I think we stopped using this due to the newer version of Solidity we are using making it obsolete. @0xcodercrane can you confirm?

0x4007 (changes requested):

0x4007:
Looks like conventional commits CI isn't passing. Be sure to setup the dependencies again in order to have conventional commits enforced locally with the pre-commit hook as well.

ethernomad:
@pavlovcik The changes you are commenting on are from #388 - you should probably comment on them over there.

This pull request currently includes the commits from #388 because it hasn't been merged yet.

ethernomad:
> Looks like conventional commits CI isn't passing. Be sure to setup the dependencies again in order to have conventional commits enforced locally with the pre-commit hook as well.

These commits were made before Conventional Commits was added to the CI.

0x4007:
Perhaps you should rebase or else we have redundant changes to review.

PR #258, opened 2022-09-13
fix: warnings

0x4007:
Got most of them but the following remain. If an easy fix feel free to take over this PR, otherwise I'll look into it later.

```
➤ YN0000: ┌ Resolution step
➤ YN0002: │ @codechecks/client@npm:0.1.12 doesn't provide typescript (pb3185), requested by ts-node
➤ YN0060: │ @ubiquity/dollar@workspace:packages/contracts/dollar provides @nomiclabs/hardhat-ethers (pe1017) with version 0.3.0-beta.13, which doesn't satisfy what @nomiclabs/hardhat-waffle requests
➤ YN0060: │ @ubiquity/ubiquistick@workspace:packages/contracts/ubiquistick provides @nomiclabs/hardhat-ethers (p86605) with version 0.3.0-beta.13, which doesn't satisfy what @nomiclabs/hardhat-waffle requests
➤ YN0002: │ typechain@npm:3.0.0 doesn't provide typescript (p4ba7f), requested by ts-essentials
➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed in 0s 421ms
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed in 0s 400ms
➤ YN0000: Done with warnings in 1s 163ms
```

PR #468, opened 2023-01-10
build: resolve yarn peer deps

### LINKED ISSUE #355: Fix Yarn Workspace Warnings
Context: https://github.com/ubiquity/ubiquity-dollar/pull/258

Creating an issue so that we don't forget about this. 

This is not intended to be released to the DevPool, but rather, handled by trusted devs. The reason why is that I think the core devs should handle this is because it may require us to remove/replace core dependencies. I think the core devs will have better judgement on this. Open to the idea of releasing into the DevPool in the future if that is a non issue.

Steveantor:
Context: https://github.com/ubiquity/ubiquity-dollar/pull/258
- update dependencies
- update .`yarnrc`

Resolves #355

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -26,3 +26,36 @@ plugins:
 
 yarnPath: .yarn/releases/yarn-3.2.3.cjs
 enableScripts: true
+
+packageExtensions:
+  "@tanstack/query-persist-client-core@4.22.0":
+    dependencies:
+      "@tanstack/query-core": "4.22.0"
+  "hardhat-watcher@^2.1.1":
+    dependencies:
+      "hardhat": "^2.0.0"
+  "@uniswap/widgets@^2.16.2":
+    dependencies:
+      "@walletconnect/ethereum-provider": "1.7.8"
+  "@uniswap/redux-multicall@^1.1.1":
+    dependencies:
+      "@ethersproject/abi": "5"
+      "@ethersproject/bignumber": "5.7.0"
+      "@ethersproject/contracts": "5.7.0"
+  "styled-components@>=5":
+    dependencies:
+      "react-is": "16.8.0"
+  "abitype@^0.1.8":
+    dependencies:
+      "typescript": "^4.7.4"
+  "eth-block-tracker@4.4.3":
+    dependencies:
+      "@babel/core": "7.4.0"
+
+logFilters:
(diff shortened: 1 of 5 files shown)


.yarnrc.yml
```diff
@@ -26,3 +26,36 @@ plugins:
...
+      "react-is": "16.8.0"
+  "abitype@^0.1.8":
+    dependencies:
+      "typescript": "^4.7.4"
+  "eth-block-tracker@4.4.3":
+    dependencies:
+      "@babel/core": "7.4.0"
+
+logFilters:
+  - pattern: "@ubiquity/dapp@workspace:packages/dapp provides react * with version * which doesn't satisfy what @uniswap/widgets and some of its descendants request"
```
0x4007, line 56: Will this filter only for Uniswap package dependents related logs then? Otherwise seems like a potentially bad workaround that can hide relevant errors down the road.

packages/contracts/package.json
```diff
@@ -57,5 +57,9 @@
       "postinstall-postinstall": false,
       "husky": false
     }
+  },
+  "devDependencies": {
+    "@types/node": "^18.11.18",
+    "typescript": "^4.7.4"
```
0x4007: Couldn't this be a later version of typescript?

.yarnrc.yml
Steveantor: It'll only resolve the `@uniswap/widgets`'s issue as it uses 

> react: ">=17.0.1"
(which is compatible with any react version equal to or above `17.0.1` (`...18.2.0...`) according to semver)

the following package also duplicates the same dependencies into `deps` and `peer-deps` (which is the root of this issue)

@pavlovcik

packages/contracts/package.json
Steveantor: the typescript version `^4.7.4` was suggested by some deps,
but it gets resolved into `4.9.4` (<5) at the end.
image
@pavlovcik

0x4007: So then shouldn't we upgrade it in the package to reflect the latest?

Steveantor: > So then shouldn't we upgrade it in the package to reflect the latest?

bumped the typescript version.

0x4007:
@Steveantor the CI got rate limited so it didn't pass. Should we make a new issue for this?

Steveantor:
> @Steveantor the CI got rate limited so it didn't pass. Should we make a new issue for this?

I think so @pavlovcik

PR #474, opened 2023-01-12
refactor: resolve RPC rate limiting

Steveantor:
- Run tests on `anvil` from now on
- Move current RPC utility function to `rpcutil.ts`
- Add `localRPC` config (using exact IP instead of _localhost_ to avoid potential resolution errors)

As it turns out, GitHub Actions use shared IPs which oftentimes get rate-limited by
the RPC providers,
If we make a bunch of calls.

I've looked around and found that the same process can spin up an anvil instance to run the tests.

Resolves #473

PR #476, opened 2023-01-12
fix: test process getting stuck

Steveantor:
Related #474

Resolves #

### DIFF
--- a/packages/contracts/scripts/runner/runner.ts
+++ b/packages/contracts/scripts/runner/runner.ts
@@ -23,9 +23,12 @@ const procFork = async () => {
     }
   });
   command.on("close", (code: number) => {
-    anvil.kill();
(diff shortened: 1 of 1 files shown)


0x4007:
@Steveantor I feel like I should just give you direct write access to `development` to hack through CI next time you have a CI related task. Alternatively we could consider making a staging repo for CI.

Steveantor:
> @Steveantor I feel like I should just give you direct write access to `development` to hack through CI next time you have a CI related task. Alternatively we could consider making a staging repo for CI.

yeah that'd be helpful @pavlovcik

0x4007:
You should have an invitation here https://github.com/orgs/ubiquity/teams/development/members

Hope that you can fix the continuous deployment!

PR #524, opened 2023-01-29
Fix/state

0x4007:
Just would like help to get these tests fixed and this should be stable:

image


zgorizzo69:
@pavlovcik the remote test are supposed to be working with a fork url 
check runner.ts for more information

PR #525, opened 2023-01-29
fix(submodules): location

0x4007:
Added [submodule "packages/contracts/lib/Uniswap/v3-periphery"]


pavlovcik:
@hashedMae I accidentally included my full codebase formatting/prettier run squashed with my attempt at fixing the slither errors, which adds noise to the diff. It would be helpful if you do it on your branch before this is merged so you can see a clean diff.

PR #531, opened 2023-01-31
feat: metapool migration script

### LINKED ISSUE #295: Migrate to V3 Metapool
crv.to is down and I can no longer find a swap UI for a v2 metapool that is currently functional. It appears that curve.fi dynamically can generate v3 metapool swap UIs, so this is a sound reason to migrate to a v3 pool officially. 

Here's the v3 pool (with no liquidity) I deployed some time ago: https://curve.fi/factory/124

        Unfortunately there is a blocking UI issue with crv.to so I need to figure out another UI for swapping on the metapools.

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquity-dollar/issues/225#issuecomment-1283003169_

hashedMae:
Resolves #295 

Old PR was based on a messy branch. Created a fresh one to clean it all up.


0x4007:
@0xcodercrane @zgorizzo69 please leave a review

hashedMae:
rebased from #554 Since y'all got the CI working

packages/contracts/src/dollar/mocks/MockBondingShareV2.sol
```diff
@@ -0,0 +1,226 @@
...
+pragma solidity ^0.8.3;
+
+import "@openzeppelin/contracts/token/ERC1155/ERC1155.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC1155/extensions/ERC1155Burnable.sol";
+import "@openzeppelin/contracts/token/ERC1155/extensions/ERC1155Pausable.sol";
+import "../core/UbiquityDollarManager.sol";
+import "../utils/SafeAddArray.sol";
+
+contract BondingShareV2 is ERC1155, ERC1155Burnable, ERC1155Pausable {
```
0x4007: We renamed this to Staking
```suggestion
contract Staking is ERC1155, ERC1155Burnable, ERC1155Pausable {
```

packages/contracts/src/dollar/mocks/MockBondingShareV2.sol
```diff
@@ -0,0 +1,226 @@
...
+        uint256 lpRewardDebt;
+        uint256 endBlock;
+        // lp remaining for a user
+        uint256 lpAmount;
+    }
+
+    UbiquityDollarManager public manager;
+    // Mapping from account to operator approvals
+    mapping(address => uint256[]) private _holderBalances;
+    mapping(uint256 => Bond) private _bonds;
```
0x4007, line 30: Bonds I think were changed to StakingToken but you should refer to the other code to see.

packages/contracts/test/dollar/Staking.t.sol
```diff
@@ -166,14 +167,16 @@ contract RemoteZeroStateTest is ZeroState {
         vm.stopPrank();
 
         assertEq(
-            bytes20(address(steak)), bytes20(address(staking.stakingFormulas()))
+            bytes20(address(steak)),
```
0x4007: I'm pretty sure I fixed this in your previous PR, you might need to rebase to a later commit. Its stake, not steak.

0x4007:
What is `s.sol` is that a standardized extension for "scripts"?

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
```diff
@@ -0,0 +1,84 @@
...
+
+			uint256 depositV3;
+
+			if(usd3Bal < uADBal){
+				depositV3 = usd3Bal;
+			} else {
+				deposit = uADBal;
+			}
+
+			USD3CRV.approve(address(v3Metapool), 0);
```
zgorizzo69: should be USD3CRVToken

zgorizzo69:
> What is `s.sol` is that a standardized extension for "scripts"?

yep

zgorizzo69:
I found that MetapoolMigrate test was not part of the CI test so I reran the CI wrokflow cos it was failing locally and now guess what it includes the MetapoolMigrate failing test :thinking:

zgorizzo69:
when I run `forge script scripts/task/dollar/MigrateMetapool.s.sol -f http://localhost:8545 --broadcast` I have an error 
`Error: 
Failed to get environment variable `PUBLIC_KEY` as type `address`: environment variable not found`
we should  get the pubkey from the private key

zgorizzo69 (changes requested):

0x4007:
> when I run forge script scripts/task/dollar/MigrateMetapool.s.sol -f http://localhost:8545 --broadcast I have an error
Error:  Failed to get environment variable PUBLIC_KEYas typeaddress: environment variable not found
we should get the pubkey from the private key

Perhaps this could also be useful to throw into our script setup. Its a personal bash function I use to generate disposable key pairs.

https://gist.github.com/pavlovcik/54fdc710064ba84e5b2bc9b5d18a9dd5

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
0x4007: Now that I'm actually looking at the script, this is known as the "3pool" we can refer to it as such: `curve3poolToken`

packages/contracts/scripts/shared/helpers/loadEnv.ts
```diff
@@ -7,7 +7,7 @@ export const loadEnv = (path: string): Env => {
   dotenv.config({ path });
   const rpcUrl = process.env.RPC_URL || "http://localhost:8545";
```
0x4007: Is this updated? I believe I merged in changes that have this default to our custom gateway `https://eth.ubq.fi/v1/mainnet`

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
```diff
@@ -0,0 +1,84 @@
...
+import "src/dollar/core/UbiquityDollarManager.sol";
+
+contract migrateFunds is Script {
+	function run() public virtual {
+
+		
+		IMetaPool v2Metapool = IMetaPool(0x20955CB69Ae1515962177D164dfC9522feef567E);
+		IMetaPool v3Metapool = IMetaPool(0x9558b18f021FC3cBa1c9B777603829A42244818b);
+		IERC20 USD3CRVToken = IERC20(0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490);
+		IERC20 uAD = IERC20(0x0F644658510c95CB46955e55D7BA9DDa9E9fBEc6);
```
0x4007: I believe I banned `uAD` via cspell, perhaps I didn't merge it yet?

@0xcodercrane and I discussed making a code styleguide we'll have to make it soon here. 

Anyways consider calling this `UbiquityDollar` as its purpose is clearer for new developers. Also we have a token symbol setter in the contract so it is possible to update the token symbol after deployment, so `uAD` could be deprecated.
```suggestion
		IERC20 UbiquityDollar = IERC20(0x0F644658510c95CB46955e55D7BA9DDa9E9fBEc6);
```

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
0x4007:
```suggestion
			curve3poolToken.approve(address(v3Metapool), 0);
```

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,164 @@
...
+
+import "forge-std/Test.sol";
+
+contract MetapoolMigrate is Test {
+    address admin = 0xefC0e701A824943b469a694aC564Aa1efF7Ab7dd;
+    UbiquityDollarManager manager =
+        UbiquityDollarManager(0x4DA97a8b831C345dBe6d16FF7432DF2b7b776d98);
+    Staking staking =
+        Staking(payable(0xC251eCD9f1bD5230823F9A0F99a44A87Ddd4CA38));
+    BondingShareV2 bonds =
```
0x4007:
```suggestion
    StakingToken stakingToken =
```

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,164 @@
...
+    IMetaPool v3Metapool =
+        IMetaPool(0x9558b18f021FC3cBa1c9B777603829A42244818b);
+
+    uint256 mainnet;
+    uint256 snapshot;
+    uint256 metaBalanceV3;
+    uint256 metaBalance;
+
+    function setUp() public {
+        mainnet = vm.createSelectFork("https://eth.ubq.fi/v1/mainnet");
```
0x4007, line 48: Seems that the fork selection is mixed. I think this is a reasonable default but in order to stay consistent with the other scripts, shouldn't it be loaded in from the `.env` and fall back to this if nothing?

0x4007 (changes requested):

packages/contracts/src/dollar/mocks/MockBondingShareV2.sol
hashedMae: This script is interacting with contracts as currently deployed. If you try to call Staking.function() on mainnet it'll revert.

packages/contracts/src/dollar/mocks/MockBondingShareV2.sol
hashedMae: This is a mock contract to interact with the contract as it's currently deployed on mainnet.

packages/contracts/test/dollar/Staking.t.sol
hashedMae: It appears this was resolved when I rebased from your PR that fixed CI

hashedMae:
> when I run `forge script scripts/task/dollar/MigrateMetapool.s.sol -f http://localhost:8545 --broadcast` I have an error `Error: Failed to get environment variable `PUBLIC_KEY`as type`address`: environment variable not found` we should get the pubkey from the private key

Solidity doesn't have a method for deriving public addresses from private keys.

packages/contracts/scripts/shared/helpers/loadEnv.ts
hashedMae: Old version

packages/contracts/test/scripts/MigrateMetapool.t.sol
hashedMae: > I found that MetapoolMigrate test was not part of the CI test so I reran the CI wrokflow cos it was failing locally and now guess what it includes the MetapoolMigrate failing test thinking

When I wrote the script initially the treasury was holding Curve V2 Metapool LP tokens so I migrated those as well. It no longer does so I've removed that part from the scripts and the tests.

hashedMae:
> > when I run `forge script scripts/task/dollar/MigrateMetapool.s.sol -f http://localhost:8545 --broadcast` I have an error `Error: Failed to get environment variable `PUBLIC_KEY`as type`address`: environment variable not found` we should get the pubkey from the private key
> 
Changed this to hardcoded admin address since this is a one off script.

packages/contracts/test/scripts/MigrateMetapool.t.sol
hashedMae:
```function setUp() public {
        
        string memory forkURL = vm.envString("RPC_URL");

        if(keccak256(abi.encodePacked(forkURL)) != keccak256(abi.encodePacked(""))){
            mainnet = vm.createSelectFork(forkURL);
        } else {
            mainnet = vm.createSelectFork("https://eth.ubq.fi/v1/mainnet");
        }
        snapshot = vm.snapshot();
    }```

File: packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol (hunk 6 above)
Comment (hashedMae) [reply packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol]: Done

**Comment by hashedMae** [comment]
> I found that MetapoolMigrate test was not part of the CI test so I reran the CI wrokflow cos it was failing locally and now guess what it includes the MetapoolMigrate failing test thinking

When I initially wrote the script, The treasury was holding V2Metapool tokens. It no longer does so I've removed those parts from the scripts and the tests.

File: packages/contracts/test/scripts/MigrateMetapool.t.sol (hunk 7 above)
Comment (hashedMae) [reply packages/contracts/test/scripts/MigrateMetapool.t.sol]: This is specifically cause of 
`BondingShareV2.Bond memory bond = bonds.getBond(tokens[0]);`
The test is comparing How many LP tokens a user would get on withdrawal pre pool migration versus post. There is no `Struct Bond` in the `StakingShare` contract so I wouldn't be able to use in order to get the this info.

File: packages/contracts/test/scripts/MigrateMetapool.t.sol (hunk 8 above)
Comment (hashedMae) [reply packages/contracts/test/scripts/MigrateMetapool.t.sol:48]: This fails in CI cause of the lack of `.env`
I preserved the pattern above by commenting it out but have reverted to just using 
```mainnet = vm.createSelectFork("https://eth.ubq.fi/v1/mainnet");```

It'd be a seperate issue but refactoring our tests to use `vm.createSelectFork()` instead of passing it as a flag in the CLI might be worthwhile.

**Comment by 0x4007** [comment]
> > I found that MetapoolMigrate test was not part of the CI test so I reran the CI wrokflow cos it was failing locally and now guess what it includes the MetapoolMigrate failing test thinking
> 
> When I initially wrote the script, The treasury was holding V2Metapool tokens. It no longer does so I've removed those parts from the scripts and the tests.

Sorry, I moved some because of a Slither error you found some months ago. But it seemed like it was not an issue so I moved it back for stakers to be able to withdraw.

File: packages/contracts/test/scripts/MigrateMetapool.t.sol (hunk 7 above)
Comment (0x4007) [reply packages/contracts/test/scripts/MigrateMetapool.t.sol]: Oh I see because its already deployed on mainnet so we can't change. Okay! Perhaps a cspell ignore on this file is in order.

**Comment by hashedMae** [comment]
It's all good. Just wanted to explain why it was failing and the changes I'd made since others were reviewing the code.

------- Original Message -------

**Comment by hashedMae** [comment]
Yeah, using the Staking contract works fine because the functions we're using here retain the same signature. I needed the old BondingShareV2 for the Struct Bond​. I only used that here in the tests though, it wasn't needed for the script itself.

------- Original Message -------

File: packages/contracts/scripts/deploy/dollar/solidityScripting/00_Constants.sol [hunk 9]
```diff
@@ -30,12 +30,11 @@ contract Constants is Script {
     uint256[] ids;
 
     string uri;
-    address admin = vm.envAddress("ADMIN_ADDRESS");
     address USDCrvToken = 0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490;
```
0x4007:
```suggestion
    address curve3poolToken = 0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490;
```

0x4007:
@0xcodercrane can you check this

cuddle-dev:
> @0xcodercrane can you check this

yep. can do

packages/contracts/test/dollar/Staking.t.sol
```diff
@@ -162,21 +162,21 @@ contract RemoteZeroStateTest is ZeroState {
             bytes20(address(staking.stakingFormulas()))
         );
         vm.startPrank(admin);
-        StakingFormulas steak = new StakingFormulas();
-        staking.setStakingFormulas(steak);
+        StakingFormulas stake = new StakingFormulas();
+        staking.setStakingFormulas(stake);
```
cuddle-dev: good typo catch!

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
```diff
@@ -0,0 +1,72 @@
...
+            payable(0xC251eCD9f1bD5230823F9A0F99a44A87Ddd4CA38)
+        );
+        UbiquityDollarManager manager = UbiquityDollarManager(
+            0x4DA97a8b831C345dBe6d16FF7432DF2b7b776d98
+        );
+
+        TWAPOracleDollar3pool twapOracle;
+
+        uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
+        address admin = 0xefC0e701A824943b469a694aC564Aa1efF7Ab7dd;
```
cuddle-dev: I am just curious if `PRIVIATE_KEY` is the key for `admin` account; 0xefC0e701A824943b469a694aC564Aa1efF7Ab7dd.

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
```diff
@@ -0,0 +1,72 @@
...
+            deposit = curve3Bal;
+        } else {
+            deposit = dollarBal;
+        }
+
+        curve3PoolToken.approve(address(v3Metapool), 0);
+        curve3PoolToken.approve(address(v3Metapool), deposit);
+
+        dollarToken.approve(address(v3Metapool), 0);
+        dollarToken.approve(address(v3Metapool), deposit);
```
cuddle-dev: same as above

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
```diff
@@ -0,0 +1,72 @@
...
+        uint256 deposit;
+
+        if (curve3Bal <= dollarBal) {
+            deposit = curve3Bal;
+        } else {
+            deposit = dollarBal;
+        }
+
+        curve3PoolToken.approve(address(v3Metapool), 0);
+        curve3PoolToken.approve(address(v3Metapool), deposit);
```
cuddle-dev: I am still not sure why we need the first transaction. could you please make me clear on it?
On `approve` function, it sets the allowance value like this `self.allowances[msg.sender][spender] = value`.
```suggestion
        curve3PoolToken.approve(address(v3Metapool), deposit);
```

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
```
cuddle-dev, line 1: I like forge script lol

cuddle-dev (changes requested):
Nice work. A couple of minor requests still exist. appreciate for doing in forge script

hashedMae:
That’s the intention. It’s the only address that’d be able to run this script. That’s why I wrote tests to ensure it functioned properly, as you’re not able to use pranks in scripts.

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
hashedMae: Zeroing out the allowance is just a good pattern to use in general. Ensures there won't be any left over allowance.

hashedMae: Any ERC20 with `safeApprove()` uses a similar pattern, zeroing out the allowance before setting it to the new value.

0x4007: I could have sworn that approve simply sets the value (does not add) this is precisely the reason why we can "revoke approval" of a contract by calling approve with the amount 0

hashedMae: It does, that's what allows this pattern to work in the first place. It can sometimes get messy if `MAX_UINT` was set for approval previously.

0x4007: I'm not a solidity developer but this looks like you're making two calls to `approve`. One to `approve(0);` one to `approve(deposit);`

The first one seems to be a waste of gas.

hashedMae: It's a standard pattern to protect from front run attacks and prevent unintended behaviour by the contract. It's the same pattern as in OpenZeppelin's SafeERC20 wrapper.

0x4007:
So what's the status on this? Was anybody else able to run the migration?

zgorizzo69:
> > when I run `forge script scripts/task/dollar/MigrateMetapool.s.sol -f http://localhost:8545 --broadcast` I have an error `Error: Failed to get environment variable `PUBLIC_KEY`as type`address`: environment variable not found` we should get the pubkey from the private key
> 
> Solidity doesn't have a method for deriving public addresses from private keys.

foundry does https://book.getfoundry.sh/cheatcodes/addr

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
zgorizzo69: - I think @hashedMae is referring to that exploit : https://docs.google.com/document/d/1YLPtQxZu1UAvO9cZ1O2RPXBbT0mooh4DYKjA_jp-RLM/edit?usp=sharing

- indeed OZ too in the safeERC20 recommends zeroing allowance to be on the safe side 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7ac4349710d3b242dd3ceea3200f1298f1c3d2b3/contracts/token/ERC20/utils/SafeERC20.sol#L37

- final note zeroing allowance will consume much less gas. Indeed `SSTORE` which is the most expensive opcode involved in the allowance will cost 22100 gas if you allow from 0 to 1 and only 5000 if you allow from 1 to 0 (only 2200 if you store 0 where it was already 0 so roughly the `SLOAD` cost)
 https://www.evm.codes/#55?fork=grayGlacier

zgorizzo69:
I ran `yarn workspace @ubiquity/contracts start:anvil` and  ` yarn workspace @ubiquity/contracts deploy:development` but it fails cos inside `01_UbiquityDollarManager.s.sol` we have  `USDCrvToken` instead of  `curve3PoolToken`

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
zgorizzo69:
```suggestion
        address admin = vm.addr(deployerPrivateKey);
        require(admin == 0xefC0e701A824943b469a694aC564Aa1efF7Ab7dd, "only admin can run the metapool migration");
```

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,178 @@
...
+
+import "src/dollar/core/UbiquityDollarManager.sol";
+import "src/dollar/Staking.sol";
+import "src/dollar/mocks/MockBondingShareV2.sol";
+import "src/dollar/interfaces/IMetaPool.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import "forge-std/Test.sol";
+
+contract MetapoolMigrate is Test {
```
zgorizzo69:
```suggestion
contract MigrateMetapool is Test {
```

to be consistent with the filename

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,178 @@
...
+    IMetaPool v2Metapool =
+        IMetaPool(0x20955CB69Ae1515962177D164dfC9522feef567E);
+    IMetaPool v3Metapool =
+        IMetaPool(0x9558b18f021FC3cBa1c9B777603829A42244818b);
+
+    uint256 mainnet;
+    uint256 snapshot;
+
+    function setUp() public {
+        /*string memory forkURL = vm.envString("RPC_URL");
```
zgorizzo69: we can clean the comments

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,178 @@
...
+
+        if (
+            keccak256(abi.encodePacked(forkURL)) !=
+            keccak256(abi.encodePacked(""))
+        ) {
+            mainnet = vm.createSelectFork(forkURL);
+        } else {
+            mainnet = vm.createSelectFork("https://eth.ubq.fi/v1/mainnet");
+        }*/
+        mainnet = vm.createSelectFork("https://eth.ubq.fi/v1/mainnet");
```
zgorizzo69, line 48: we can add comment explaining what we are trying to do with this line and extract the URL from the env can be cool

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,178 @@
...
+            "Amount of UbiquityDollar Tokens in Curve Metapool V2: ",
+            v2dollarBal
+        );
+        console.log(
+            "Amount of Curve3 Tokens in Curve Metapool V2: ",
+            v2curve3Bal
+        );
+        console.log(
+            "Amount of UbiquityDollar Tokens in Curve Metapool V3: ",
+            v3dollarBal
```
zgorizzo69: don't we need some require statement ?

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,178 @@
...
+        curve3Token.approve(address(v3Metapool), 0);
+        curve3Token.approve(address(v3Metapool), deposit);
+
+        dollarToken.approve(address(v3Metapool), 0);
+        dollarToken.approve(address(v3Metapool), deposit);
+
+        v3Metapool.add_liquidity([deposit, deposit], 0, address(staking));
+        metaBalanceV3 = v3Metapool.balanceOf(address(staking));
+
+        require(metaBalanceV3 > 0);
```
zgorizzo69: We need more than one assertion for the tests.
non exhaustive list of what we want to make sure of : 
 - no tokens are lost during the migrations 
- LP are calculated right 
- LP holders (v2 and v3) are correct

zgorizzo69 (changes requested):

packages/contracts/test/scripts/MigrateMetapool.t.sol
hashedMae: To extract from env we would have to refactor the CI test run to create an env file. I have it hardcoded right now for simplicity but that would be useful in the future.

packages/contracts/test/scripts/MigrateMetapool.t.sol
hashedMae: This is an internal function that contains the Migration execution. I can flesh those out in the tests themselves.

0x4007:
@zgorizzo69 I will delegate this PR to you because 1. I am not able to easily test and 2. My only contributions are pretty insignificant. Just related to refactoring variable names.

0x4007 (commented):
I see a handful of identifiers with shortened "Balance" 
e.g. `v3DollarBal`

Please write the full word, Balance, in order to increase code clarity. Generally speaking, don't shorten words in identifiers. Besides, I'm pretty sure that shouldn't pass cspell check.

zgorizzo69:
> I ran `yarn workspace @ubiquity/contracts start:anvil` and ` yarn workspace @ubiquity/contracts deploy:development` but it fails cos inside `01_UbiquityDollarManager.s.sol` we have `USDCrvToken` instead of `curve3PoolToken`

@hashedMae ping me when this is fixed and when you rebase so that there is no more conflict so that I can do a last review

hashedMae:
> > I ran `yarn workspace @ubiquity/contracts start:anvil` and ` yarn workspace @ubiquity/contracts deploy:development` but it fails cos inside `01_UbiquityDollarManager.s.sol` we have `USDCrvToken` instead of `curve3PoolToken`
> 
> @hashedMae ping me when this is fixed and when you rebase so that there is no more conflict so that I can do a last review

@zgorizzo69 ping

zgorizzo69:
> > > I ran `yarn workspace @ubiquity/contracts start:anvil` and ` yarn workspace @ubiquity/contracts deploy:development` but it fails cos inside `01_UbiquityDollarManager.s.sol` we have `USDCrvToken` instead of `curve3PoolToken`
> > 
> > 
> > @hashedMae ping me when this is fixed and when you rebase so that there is no more conflict so that I can do a last review
> 
> @zgorizzo69 ping

cspell is failing

0x4007:
> > > > I ran `yarn workspace @ubiquity/contracts start:anvil` and ` yarn workspace @ubiquity/contracts deploy:development` but it fails cos inside `01_UbiquityDollarManager.s.sol` we have `USDCrvToken` instead of `curve3PoolToken`
> > > 
> > > 
> > > @hashedMae ping me when this is fixed and when you rebase so that there is no more conflict so that I can do a last review
> > 
> > 
> > @zgorizzo69 ping
> 
> cspell is failing

I'm gonna fix that now

0x4007:
Hey @hashedMae I got you on https://github.com/ubiquity/ubiquity-dollar/pull/583

Please verify I didn't break something but I am reasonably confident that this should be fine.

When reviewing my pull request, please focus only on the files you're changing in this PR. In my cspell fixes I changed a lot of UI code, for example (which you should ignore!)

0x4007:
Update: https://github.com/ubiquity/ubiquity-dollar/pull/584

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,249 @@
...
+            "Curve Metapool V3 UbiquityDollar balance after withdrawal: ",
+            v3DollarBalance
+        );
+        console.log(
+            "Curve Metapool V3 Curve3 balance after withdrawal: ",
+            v3Curve3Balance
+        );
+
+        assertGe(v3Balance, v2Balance);
+        assertLt(userDollarV3, userDollarV2);
```
zgorizzo69: there is a difference of about 3907 dollar for the user between when withdrawing with v2 and v3
I guess that is because when migrating to v3 as v2 pool is unbalanced we don't have the same dollar amount in the v3 pool
but we have to make sure that  the user is not loosing one penny so after migrating user should withdraw from V2 AND v3 and we should het the same amount as if there were no migration and user withdraw all from v2
In short we should have strong assertion in this test "Greater than" and "Less than" are not enough to really make sure nothing is lost for the users 
Also it is worth checking that all dollar in pools stay the same e.g:
- in V2 no migration  
V2 pool's dollar balance =  X dollars
- migration to v3
V2 pool's dollar balance =  Y dollars
V3 pool's dollar balance =  Z dollars
X = Y +Z

zgorizzo69: image

zgorizzo69 (changes requested):

packages/contracts/test/scripts/MigrateMetapool.t.sol
hashedMae: Only way I can think of to resolve that is do a single sided liquidity add after launching the TWAPOracle for v3 pool

hashedMae:
Major issue I just caught, script wasn't setting the new `stableSwapMetapoolAddress` after migrating funds and deploying a new `TWAPOracle`

packages/contracts/test/scripts/MigrateMetapool.t.sol
zgorizzo69: if we withdraw everything from V2 and V3 (so after migration) we should get the same amount of dollar than if we withdraw everything from v2 (without migration) right ?

hashedMae: Screenshot from 2023-02-27 12-03-49
I was able to get the values a lot closer but not equal. I believe this is cause the V3 Metapool is minting more LP Tokens than we burned for V2 and we don't control the entirety of the V2 Metapool so there will always be some tokens leftover in the old pool.

zgorizzo69: Ok lp tokens amount make less sense than the coins behind. Can you check the exact amount of dollar/3crv behind?

hashedMae: I added that in the logs earlier as well. 
`Amount of UbiquityDollar Tokens in Curve Metapool V2 after migration:  169833975952106434827`
  `Amount of Curve3 Tokens in Curve Metapool V2 after migration:  142863822020848582307`

hashedMae: The X=Y+Z checks out. `assertEq()` for those is at lines 121 and 124
image

packages/contracts/test/diamond/DiamondTestSetup.sol
```diff
@@ -404,6 +404,13 @@ abstract contract DiamondSetup is DiamondTestHelper {
 
         DiamondInit.Args memory initArgs = DiamondInit.Args({
             admin: admin,
+            /*
+            // cspell: disable-next-line
+            dollarName: "Ubiquity Algorithmic Dollar",
+            // cspell: disable-next-line
+            dollarSymbol: "uAD",
+            dollarDecimals: 18,
+            */
```
hashedMae: @pavlovcik These threw an error on compile cause the `Args` struct in `DiamondInit` only contains 6 values. I commented them out rather than removing them cause it was unclear if they should be added to the struct or not.

packages/contracts/test/scripts/MigrateMetapool.t.sol
hashedMae: In the withdrawal comparison between pre and post migration, there's only a difference of ~.20 in `uAD` but `curve3` is ~143. This is fairly significant and I don't think it can be remedied without adding additional `curve3` to the pool.

hashedMae:
So the underlying issue is that the uAD:3CRV pool that has liquidity in it isn't showing up in any interfaces. Curve has a V2 registry and a V3 registry, I'm guessing that it's checking the V3 registry, finding the empty pool there, and then not bothering to check the V2 registry. 
We can't migrate the funds without lowering the value of existing Bonds. 
A better course of action may be to reach out to Curve for a solution. We can't delete the V3 Metapool but it looks like we were hasty in deploying it to start.

packages/contracts/test/scripts/MigrateMetapool.t.sol
hashedMae: I was looking at the wrong numbers. Difference in withdrawal is ~.2 less for each token after migration. Comes out to to 0.002% less.

packages/contracts/src/dollar/Staking.sol
```diff
@@ -162,10 +162,10 @@ contract Staking is IStaking, CollectableDust, Pausable {
         migrating = _migrating;
     }
 
-    /// @dev uADPriceReset remove uAD unilaterally from the curve LP share sitting inside
-    ///      the staking contract and send the uAD received to the treasury.
-    ///      This will have the immediate effect of pushing the uAD price HIGHER
-    /// @param amount of LP token to be removed for uAD
+    /// @dev uADPriceReset remove Dollars unilaterally from the curve LP share sitting inside
```
0x4007: `uADPriceReset` looks like I missed one.

packages/contracts/src/dollar/mocks/MockShareV1.sol
```diff
@@ -6,5 +6,7 @@ import "src/dollar/core/UbiquityDollarManager.sol";
 
 contract BondingShare is ERC1155Ubiquity {
```
0x4007: This should also be MockBondingShare I guess

0x4007:
Just checked from when you merged in my cspell changes to HEAD. Code style looks fine. I guess the only questionable identifier is that you were referring to "Curve3" Tokens I'm looking for a line number now

Checked: https://github.com/ubiquity/ubiquity-dollar/compare/0d2c45d23f706f4a42b27fc296de0c8f891a77f8..fix/295-migrate-to-v3-metapool-b

packages/contracts/test/scripts/MigrateMetapool.t.sol
```diff
@@ -0,0 +1,261 @@
...
+        console.log(
+            "Amount of Ubiquity Dollar Tokens in Curve Metapool V2 after migration: ",
+            v2DollarPostBalance
+        );
+        console.log(
+            "Amount of Ubiquity Dollar Tokens in Curve Metapool V3: ",
+            v3DollarBalance
+        );
+        console.log(
+            "Amount of Curve3 Tokens in Curve Metapool V2 before migration: ",
```
0x4007, line 98: For example, I thought that these should be called something like:

- curve3PoolToken
- threePoolToken

0x4007 (approved):

packages/contracts/scripts/task/dollar/MigrateMetapool.s.sol
```diff
@@ -0,0 +1,106 @@
...
+
+contract migrateFunds is Script {
+    function run() public virtual {
+        IMetaPool v2Metapool = IMetaPool(
+            0x20955CB69Ae1515962177D164dfC9522feef567E
+        );
+        IMetaPool v3Metapool = IMetaPool(
+            0x9558b18f021FC3cBa1c9B777603829A42244818b
+        );
+        IERC20 curve3PoolToken = IERC20(
```
hashedMae, line 20: @pavlovcik I did

zgorizzo69 (approved):

PR #538, opened 2023-02-02
fix: anvil ts mnemonic

hashedMae:
Resolves #

quick fix cause anvil.ts was hardcoded to use one mnemonic instead of pulling from .env


rndquu:
@pavlovcik merged the `development` branch so that the "build and test" could pass

0x4007:
For the future: I think rebasing is cleaner but if this solves the problem for now then ok!

PR #540, opened 2023-02-05
Update NFT case to camelcase

Steveantor:
Updating from [refactor/general-cleanup](https://github.com/ubiquity/ubiquity-dollar/tree/refactor/general-cleanup)
Resolves #

PR #553, opened 2023-02-07
Refactor/prettier all

0x4007:
Ran prettier on everything.


Steveantor:
I think this needs to be rebased since #540 changed a lot

0x4007:
> I think this needs to be rebased since #540 changed a lot

Prettier can simply be re-ran from development HEAD.

PR #558, opened 2023-02-09
refactor(NFT): rename NFT to Nft in method names

0x4007:
Camelcase NFT to Nft


Steveantor:
wasn't it already done in #540?

0x4007:
> wasn't it already done in #540?

Yes but I was working on a trunk with a ton of general changes and then opened many PRs from it. Basically it was a large divergence from the codebase with its base being from before your PR.

PR #565, opened 2023-02-12
Refactor Cloudflare Pages Build

Steveantor:
**Project Setup (GitHub secrets):**

**CLOUDFLARE_ACCOUNT_ID =**
> https://dash.cloudflare.com/***/pages
> https://dash.cloudflare.com/abcd1234/pages
> (Here `abcd1234` is your account ID)

**CLOUDFLARE_API_TOKEN =**

> https://dash.cloudflare.com/profile/api-tokens > Create Token > API token templates > Edit Cloudflare Workers > Use Template
> 
> > Account Resources = All Resources
> 
> > Zone Resources = All Zones
> 
> (Detailed Instructions: https://developers.cloudflare.com/workers/wrangler/ci-cd/)

**CLOUDFLARE_ASSET_DIRECTORY =** 

> packages/dapp/dist

**CLOUDFLARE_PROJECT_NAME =**

> npm install -g wrangler
> wrangler login
> wrangler pages project create

_Cloudflare Pages_ couldn't build precompiled foundry package since it relies on
`GLIBC_2.25` and above which _Cloudflare Pages v1_ couldn't provide,
hence we're using _wrangler_ action.

Resolves #529

PR #568, opened 2023-02-14
feat(cloudflare): faster foundry setup fix

0x4007:
Refactors #565


0x4007:
6 minutes down from 20!

PR #576, opened 2023-02-17
Bump Node Version to 18

Steveantor:
Resolves #


.github/workflows/yarn-audit.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: "18"
+          node-version: "v18.14.1"
```
0x4007: You sure this is correct syntax

Steveantor: "v" was accidentally added, updated.

PR #582, opened 2023-02-24
fix(diamond): remove token and add more roles to the manager

### LINKED ISSUE #577: Extract dollar form diamond and improve access control granularity for all tokens
we should extract the dollar ERC20 from the diamond and the access control from the diamond should be injected in the constructor
- add the dollar address inside the manager facet
- adapt all the facets and tests that reference it 
- create new burn and mint  roles for each of the ubiquity token

the dollar should be behind a UUPS proxy in order to be able to upgrade it ( we should have the tests improved to assess the upgradability)

zgorizzo69:
Resolves #577


packages/contracts/src/diamond/libraries/LibCreditNFTManager.sol
```diff
@@ -153,10 +152,10 @@ library LibCreditNftManager {
         );
 
         // we burn user's dollars.
-        LibDollar.burn(msg.sender, amount);
+        IERC20Ubiquity(s.dollarTokenAddress).burnFrom(msg.sender, amount);
         // mint Credit
         UbiquityCreditToken creditToken = UbiquityCreditToken(
-            LibAppStorage.appStorage().creditTokenAddress
+            s.creditTokenAddress
```
0x4007: Not a fan of one letter identifiers. Why not just call it storage or store

packages/contracts/src/diamond/token/ERC20UbiquityForDiamond.sol
```diff
@@ -28,35 +27,19 @@ contract ERC20UbiquityForDiamond is
...
-        );
-        _;
-    }
-
+    // modifiers
     modifier onlyPauser() {
         require(
             accessCtrl.hasRole(PAUSER_ROLE, msg.sender),
-            "Governance token: not pauser"
+            "ERC20: not pauser"
```
0x4007:
```suggestion
            "ERC20Ubiquity: not pauser"
```

packages/contracts/src/diamond/libraries/LibCreditNFTManager.sol
zgorizzo69: storage can't be used as a variable name as it is a keyword in solidity  
will rename to store then

packages/contracts/src/diamond/token/ERC20UbiquityForDiamond.sol
zgorizzo69: ok done

0x4007:
Looks good to me!

PR #583, opened 2023-02-25
Fix/295 migrate to v3 metapool cspell

0x4007:
Made a lot of small changes to fix cspell, this should sync your branch


0x4007:
# WARNING!

@zgorizzo69 I removed the v1 staking related stuff as it is deprecated. https://github.com/ubiquity/ubiquity-dollar/pull/583/commits/481518616b7d2b8985d1ed5b05a792235e553469

This also means that I removed the migration features. I feel like we could easily add them back in by manually checking an earlier commit? I have mixed feelings on this approach, but it was the fastest way for my naive self to remove all the deprecated (v1) code.

packages/contracts/test/helpers/LiveTestHelper.sol
```diff
@@ -103,19 +98,6 @@ contract LiveTestHelper is Test {
...
-        manager.setStakingContractAddress(address(stakingV1));
-        manager.grantRole(
-            manager.GOVERNANCE_TOKEN_MINTER_ROLE(),
-            address(stakingV1)
-        );
-        manager.grantRole(
-            manager.GOVERNANCE_TOKEN_MINTER_ROLE(),
-            address(stakingShareV1)
-        );
-
```
0x4007, line 118: This looks like it only tested v1? That doesn't seem right at all!

0x4007:
I just redid everything very carefully on https://github.com/ubiquity/ubiquity-dollar/pull/584

PR #584, opened 2023-02-25
Fix/295 migrate to v3 metapool cspell tests

0x4007:
Second attempt I was much more careful about the tests (and not removing anything unnecessary!)


packages/contracts/src/dollar/mocks/MockStakingShare.sol
```diff
@@ -8,7 +8,7 @@ import "@openzeppelin/contracts/token/ERC1155/extensions/ERC1155Pausable.sol";
 import "../core/UbiquityDollarManager.sol";
 import "../utils/SafeAddArray.sol";
 
-contract BondingShareV2 is ERC1155, ERC1155Burnable, ERC1155Pausable {
```
0x4007, line 11: This didn't have Mock as a prefix, the only one. I assume that was a mistake.

packages/contracts/src/dollar/mocks/MockBondingV1.sol
```diff
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: MIT
+// cspell: disable // disabled because all of bonding v1 is deprecated
```
0x4007, line 2: This file might need to be removed?

hashedMae: This was for testing the migration from V1 to V2 when I was doing the investigation.

hashedMae (approved):

0x4007:
I'm feeling kind of sick today but would you mind handling those merge conflicts and merging to your branch soon? It will help with the spelling errors! @hashedMae

PR #586, opened 2023-02-27
chore(deployment): fix diamond deploy script

### LINKED ISSUE #585: `DiamondInAll.ts` broken syntax with merge
> @zgorizzo69 btw I'm running prettier on everything and noticed that this commit super broke DiamondInAll.ts you should check again.

https://github.com/ubiquity/ubiquity-dollar/commit/36de2d0c274b13c86334c86b7c8285d9aa5729c1#r102057720

image

zgorizzo69:
Resolves #585

`DiamondInit` is a special contract that is kind of a constructor for the diamond

As it is now the diamond Init contract is way to gas heavy to be deployed 
in this pr I removed some token's deployment that were happening inside the diamond init.

Note that  Access control is inside the diamond  and each token has the diamond address to be able to call access control logic . 

So we have to first deploy the diamond and then the tokens because token's constructor needs the diamond address.

### DIFF
--- a/packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
+++ b/packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
@@ -5,8 +5,7 @@ export const FacetCutAction = { Add: 0, Replace: 1, Remove: 2 };
 
 export async function getSelectorsFromFacet(contractName: string, artifactFolderPath = "../../../out") {
   const contractFilePath = path.join(artifactFolderPath, `${contractName}.sol`, `${contractName}.json`);
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const contractArtifact = require(contractFilePath);
+  const contractArtifact = await import(contractFilePath);
   const abi = contractArtifact.abi;
   const bytecode = contractArtifact.bytecode;
   const target = new ethers.ContractFactory(abi, bytecode);
@@ -24,10 +23,9 @@ export async function getSelectorsFromFacet(contractName: string, artifactFolder
   return selectors;
 }
 
-export function getContractInstance(contractName: string, account?: ethers.Signer, artifactFolderPath = "../../../out") {
+export async function getContractInstance(contractName: string, account?: ethers.Signer, artifactFolderPath = "../../../out") {
   const contractFilePath = path.join(artifactFolderPath, `${contractName}.sol`, `${contractName}.json`);
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const contractArtifact = require(contractFilePath);
+  const contractArtifact = await import(contractFilePath);
   const abi = contractArtifact.abi;
   const bytecode = contractArtifact.bytecode;
   let target;
--- a/packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
+++ b/packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
@@ -24,7 +24,7 @@ export async function getSelectorsFromFacet(contractName: string, artifactFolder
   return selectors;
 }
 
-export function getContractInstance(contractName: string, account?: ethers.Signer | undefined, artifactFolderPath = "../../../out") {
+export function getContractInstance(contractName: string, account?: ethers.Signer, artifactFolderPath = "../../../out") {
   const contractFilePath = path.join(artifactFolderPath, `${contractName}.sol`, `${contractName}.json`);
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const contractArtifact = require(contractFilePath);
--- a/packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
+++ b/packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
@@ -5,15 +5,15 @@ export const FacetCutAction = { Add: 0, Replace: 1, Remove: 2 };
 
 export async function getSelectorsFromFacet(contractName: string, artifactFolderPath = "../../../out") {
   const contractFilePath = path.join(artifactFolderPath, `${contractName}.sol`, `${contractName}.json`);
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const contractArtifact = require(contractFilePath);
   const abi = contractArtifact.abi;
   const bytecode = contractArtifact.bytecode;
   const target = new ethers.ContractFactory(abi, bytecode);
   const signatures = Object.keys(target.interface.functions);
 
-  const selectors = signatures.reduce((acc, val) => {
+  const selectors = signatures.reduce((acc: string[], val) => {
     if (val !== "init(bytes)") {
-      // @ts-ignore
       acc.push(target.interface.getSighash(val));
     }
     return acc;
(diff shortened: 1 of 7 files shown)


packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
```diff
@@ -24,8 +24,9 @@ export async function getSelectorsFromFacet(contractName: string, artifactFolder
   return selectors;
 }
 
-export function getContractInstance(contractName: string, account?: any, artifactFolderPath = "../../../out") {
+export function getContractInstance(contractName: string, account?: ethers.Signer | undefined, artifactFolderPath = "../../../out") {
```
0x4007:
```suggestion
export function getContractInstance(contractName: string, account?: ethers.Signer, artifactFolderPath = "../../../out") {
```

Code smell. Consider checking the parameter before or after its passed in.

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
```diff
@@ -10,258 +10,255 @@ import { Networks } from "../../shared/constants/networks";
...
+  name: string;
+  network: string;
+  rpcUrl: string;
+  privateKey: string;
+  contractInstance: string;
+  constructorArguments?: [DiamondArgs, cutType[]];
+  etherscanApiKey?: string;
 };
 
 const create = async (args: ForgeArguments): Promise<{ result: DeploymentResult | undefined; stderr: string }> => {
```
0x4007, line 40:
```suggestion
const create = async (args: ForgeArguments): Promise<{ result: DeploymentResult; stderr: string }> => {
```

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
```diff
@@ -10,258 +10,255 @@ import { Networks } from "../../shared/constants/networks";
...
-        const diamondArgs = args.constructorArguments[0];
-        const diamondCut = args.constructorArguments[1];
+  let flattenConstructorArgs = `"(`;
+  let prepareCmd: string;
+  if (args.constructorArguments) {
+    const diamondArgs = args.constructorArguments[0];
+    const diamondCut = args.constructorArguments[1];
 
-    flattenConstructorArgs += `${diamondArgs.owner},${diamondArgs.init},${diamondArgs.initCalldata})\" \"[`;
+    flattenConstructorArgs += `${diamondArgs.owner},${diamondArgs.init},${diamondArgs.initCalldata})" "[`;
```
0x4007: When concatenating strings in JavaScript, I highly recommend not using `+=` because if something else other than a string slips in it messes everything up. For quick concatenations, I use `String.concat(String)`. For long concatenatons, consider adding the strings into an array, and then calling Array.join(`,`);

```javascript
[ flattenConstructorArgs, diamondArgs.owner, diamondArgs.init, diamondArgs.initCalldata ].join(',')
```

Something like that.

packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
```diff
@@ -5,15 +5,15 @@ export const FacetCutAction = { Add: 0, Replace: 1, Remove: 2 };
 
 export async function getSelectorsFromFacet(contractName: string, artifactFolderPath = "../../../out") {
   const contractFilePath = path.join(artifactFolderPath, `${contractName}.sol`, `${contractName}.json`);
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const contractArtifact = require(contractFilePath);
```
0x4007: Why not use import?

packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
```diff
@@ -24,8 +24,9 @@ export async function getSelectorsFromFacet(contractName: string, artifactFolder
   return selectors;
 }
 
-export function getContractInstance(contractName: string, account?: any, artifactFolderPath = "../../../out") {
+export function getContractInstance(contractName: string, account?: ethers.Signer | undefined, artifactFolderPath = "../../../out") {
   const contractFilePath = path.join(artifactFolderPath, `${contractName}.sol`, `${contractName}.json`);
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const contractArtifact = require(contractFilePath);
```
0x4007: await import

packages/contracts/scripts/deploy/manager/PrintSelectors.ts
```diff
@@ -12,15 +12,15 @@ if (args.length != 1) {
 
 async function printSelectors(contractName: string, artifactFolderPath = "../../../out") {
   const contractFilePath = path.join(artifactFolderPath, `${contractName}.sol`, `${contractName}.json`);
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
   const contractArtifact = require(contractFilePath);
```
0x4007: await import

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
zgorizzo69: I was just fixing a lint error but I agree and would even say don't use concat just use tild `

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
zgorizzo69: if we stdout is null then result will be undefined check line 95 so we need that return type

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
zgorizzo69: done

packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
zgorizzo69: done

packages/contracts/scripts/deploy/manager/DiamondDeployHelper.ts
zgorizzo69: done

packages/contracts/scripts/deploy/manager/PrintSelectors.ts
zgorizzo69: done refactoring old code with import !

packages/contracts/scripts/deploy/manager/DiamondInAll.ts
0x4007: I'm unsure what you mean by line 95 its just a regex.

zgorizzo69: ok 94 ..the point is we need to be able to return undefined

0x4007: Doesn't it make more sense to return a non zero exit code, or at least null? That signals intent unlike undefined.

zgorizzo69: It makes sense to return undefined as the variable result is not even set if stdout is null

0x4007:
I guess we need to upgrade our conventional commits CI to support GitHub webapp commit messages.

PR #604, opened 2023-03-09
feat: base settings for smtchecker and prebuild script for exporting …

### LINKED ISSUE #578: SMTChecker
We should do the initial setup of this https://docs.soliditylang.org/en/v0.8.19/smtchecker.html

I presume there will be modifications we need to do to the code to get this to work, those can be in separate bounties after this is feature is set up.

https://twitter.com/paulrberg/status/1627654614811115520?s=46&t=lQUXMD9A0xd14FiMGl8IEQ

d-roak:
Smtchecker and prebuild script for extracting contracts.

@pavlovcik two questions about the preferred solution.
1. I added the prebuild script on the `npm run build` to ensure that we always have an updated listing of the contracts. Would you like for me to separate it?
2. I configured the smtchecker with some targets, which greatly downgrade the build time. Do you want to keep it as it is?

Resolves #578

### DIFF
--- a/packages/contracts/foundry.toml
+++ b/packages/contracts/foundry.toml
@@ -108,11 +108,3 @@ targets = [
 'src/ubiquistick/UbiquiStick.sol' = [ 'UbiquiStick' ]
 'src/ubiquistick/UbiquiStickSale.sol' = [ 'UbiquiStickSale' ]
 
-
-
-
-
-
-
-
-
--- a/packages/contracts/foundry.toml
+++ b/packages/contracts/foundry.toml
@@ -23,3 +23,96 @@ quote_style = 'double'
 
 [rpc_endpoints]
 mainnet = "https://eth.ubq.fi/v1/mainnet"
+
+[profile.default.model_checker]
+engine = 'chc'
+timeout = 10_000
+targets = [
+    'assert',
+    'constantCondition',
+    'divByZero',
+    'outOfBounds',
+    'overflow',
+    'underflow',
+]
+
+[profile.default.model_checker.contracts]
+'src/diamond/Diamond.sol' = [ 'Diamond' ]
+'src/diamond/access/AccessControlInternal.sol' = [ 'AccessControlInternal' ]
+'src/diamond/facets/AccessControlFacet.sol' = [ 'AccessControlFacet' ]
+'src/diamond/facets/ChefFacet.sol' = [ 'ChefFacet' ]
+'src/diamond/facets/CollectableDustFacet.sol' = [ 'CollectableDustFacet' ]
+'src/diamond/facets/CreditNFTManagerFacet.sol' = [ 'CreditNftManagerFacet' ]
+'src/diamond/facets/CreditNFTRedemptionCalculatorFacet.sol' = [ 'CreditNftRedemptionCalculatorFacet' ]
+'src/diamond/facets/CreditRedemptionCalculatorFacet.sol' = [ 'CreditRedemptionCalculatorFacet' ]
+'src/diamond/facets/DiamondCutFacet.sol' = [ 'DiamondCutFacet' ]
+'src/diamond/facets/DiamondLoupeFacet.sol' = [ 'DiamondLoupeFacet' ]
+'src/diamond/facets/DollarMintCalculatorFacet.sol' = [ 'DollarMintCalculatorFacet' ]
+'src/diamond/facets/DollarMintExcessFacet.sol' = [ 'DollarMintExcessFacet' ]
+'src/diamond/facets/ManagerFacet.sol' = [ 'ManagerFacet' ]
+'src/diamond/facets/OwnershipFacet.sol' = [ 'OwnershipFacet' ]
+'src/diamond/facets/StakingFacet.sol' = [ 'StakingFacet' ]
+'src/diamond/facets/StakingFormulasFacet.sol' = [ 'StakingFormulasFacet' ]
+'src/diamond/facets/TWAPOracleDollar3poolFacet.sol' = [ 'TWAPOracleDollar3poolFacet' ]
+'src/diamond/libraries/LibAppStorage.sol' = [ 'Modifiers' ]
+'src/diamond/mocks/MockCurveFactory.sol' = [ 'MockCurveFactory' ]
+'src/diamond/mocks/MockShareV1.sol' = [ 'BondingShareForDiamond' ]
+'src/diamond/token/ERC1155UbiquityForDiamond.sol' = [ 'ERC1155UbiquityForDiamond' ]
+'src/diamond/token/ERC20UbiquityForDiamond.sol' = [ 'ERC20UbiquityForDiamond' ]
+'src/diamond/token/StakingShareForDiamond.sol' = [ 'StakingShareForDiamond' ]
+'src/diamond/token/UbiquityCreditTokenForDiamond.sol' = [ 'UbiquityCreditTokenForDiamond' ]
+'src/diamond/token/UbiquityDollarTokenForDiamond.sol' = [ 'UbiquityDollarTokenForDiamond' ]
+'src/diamond/token/UbiquityGovernanceTokenForDiamond.sol' = [ 'UbiquityGovernanceTokenForDiamond' ]
+'src/diamond/upgradeInitializers/DiamondInit.sol' = [ 'DiamondInit' ]
+'src/dollar/DirectGovernanceFarmer.sol' = [ 'DirectGovernanceFarmer' ]
+'src/dollar/ERC1155SetUri/ERC1155BurnableSetUri.sol' = [ 'ERC1155BurnableSetUri' ]
+'src/dollar/ERC1155SetUri/ERC1155PausableSetUri.sol' = [ 'ERC1155PausableSetUri' ]
+'src/dollar/ERC1155SetUri/ERC1155SetUri.sol' = [ 'ERC1155SetUri' ]
+'src/dollar/ERC1155Ubiquity.sol' = [ 'ERC1155Ubiquity' ]
+'src/dollar/ERC20Ubiquity.sol' = [ 'ERC20Ubiquity' ]
+'src/dollar/Staking.sol' = [ 'Staking' ]
+'src/dollar/StakingFormulas.sol' = [ 'StakingFormulas' ]
+'src/dollar/StakingShare.sol' = [ 'StakingShare' ]
+'src/dollar/SushiSwapPool.sol' = [ 'SushiSwapPool' ]
+'src/dollar/UbiquityChef.sol' = [ 'UbiquityChef' ]
+'src/dollar/UbiquityFormulas.sol' = [ 'UbiquityFormulas' ]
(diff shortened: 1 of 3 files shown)


0x4007:
Delegating your questions to @rndquu

rndquu:
@jpldcarvalho 

Hi, thank you for the PR

> I added the prebuild script on the npm run build to ensure that we always have an updated listing of the contracts. Would you like for me to separate it?

Can't we use some regexp here https://github.com/ubiquity/ubiquity-dollar/pull/604/files#diff-5084a3d2809ef6c86199f46d1d985964c92c3a98247da067d6cb04b18b0a40a9R39 in order not to update the list of contracts on each build? Are all contracts analysed if we, for example, remove the `profile.default.model_checker.contracts` setting?

> I configured the smtchecker with some targets, which greatly downgrade the build time. Do you want to keep it as it is?

So what is the downgraded build time? I can't see any successful builds in this PR so it is hard to say whether it is a significant change.

d-roak:
> Can't we use some regexp here https://github.com/ubiquity/ubiquity-dollar/pull/604/files#diff-5084a3d2809ef6c86199f46d1d985964c92c3a98247da067d6cb04b18b0a40a9R39 in order not to update the list of contracts on each build? Are all contracts analysed if we, for example, remove the `profile.default.model_checker.contracts` setting?

Sadly, it is not possible to use a regex there. If we remove the `profile.default.model_checker.contracts` setting, it will be the same as we have now ([source](https://docs.soliditylang.org/en/v0.8.17/smtchecker.html#verified-contracts)). I was not aware of that.

> So what is the downgraded build time? I can't see any successful builds in this PR so it is hard to say whether it is a significant change.

I only tested on my machine (in the CI it seems like it times out). This occurs because of the mix `model_checker.contracts` and `model_checker.targets`, and the contracts have the most impact. E.g. results:
6 targets 6 contracts - 3m 43s
6 targets 3 contracts - 2m 17s
3 targets 3 contracts - 2m 8s

With that in mind, I still think the script is useful to gather all the contracts/files and list them, but it should be removed from the `npm run build` and it should be filtered somehow,e.g. only scanning the [core contracts](https://dao.ubq.fi/smart-contracts).
Any ideas?

0x4007:
> in the CI it seems like it times out

GitHub Actions runners can run for six hours before timing out as a heads up.

packages/contracts/package.json
```diff
@@ -33,7 +33,7 @@
     "clean": "run-p clean:*",
     "clean:yarn": "yarn clean",
     "clean:forge": "forge clean",
-    "build": "forge build",
+    "build": "tsx ./scripts/prebuild/export-smtchecker-contracts.ts && forge build",
```
0x4007, line 36:
```suggestion
    "build": "run-s build:*",
    "build:smt": "tsx ./scripts/prebuild/export-smtchecker-contracts.ts",
    "build:compile": "forge build",
```

`run-s` should run those in series. You can also parallelize with `run-p`

I didn't actually test this here, I only worked on it in this comment, so you might want to verify I have no syntax errors etc before moving on.

packages/contracts/scripts/prebuild/export-smtchecker-contracts.ts
```diff
@@ -0,0 +1,46 @@
+import fs from "fs";
+import path from "path";
+
+const foundry_file = path.join(__dirname, "../../foundry.toml");
```
0x4007, line 4: Relative imports are brittle. I don't have a better suggestion for now other than recursively looping up until the `foundry.toml` file is found.

d-roak: I followed what you had [here](https://github.com/ubiquity/ubiquity-dollar/blob/372f0eb86e467af745d8ed5a1c9b615813c43a8b/packages/contracts/scripts/deploy/export.ts#L5) and some other places. Please confirm if you want me to change it.

0x4007: Ideally, yes. The further away the relative import is, the more shaky it is.

packages/contracts/package.json
0x4007: You can also specify the order via `run-s build:smt build:compile`

d-roak: You don't have the `npm-run-all` dependency on that workspace, do you want me to add it? Apart from that, you don't have any tasks, you can only run `run-p build` or `run-s build`

packages/contracts/scripts/prebuild/export-smtchecker-contracts.ts
d-roak: What should be the limit of the search? If in some cases, there is no `foundry.toml`, we should prevent it from iterating until root, right?

packages/contracts/package.json
0x4007: > don't have the `npm-run-all` dependency on that workspace

That's strange I didn't notice that before. Yeah I think adding it is fine!

> you don't have any tasks

My code suggestion is to add those two sub tasks so you can run them both with `run-s build:smt build:compile` or in parallel if appropriate.

packages/contracts/scripts/prebuild/export-smtchecker-contracts.ts
0x4007: I'm not entirely sure. @rndquu perhaps you have an opinion here. I would default to something like
```javascript
const npmRoot = process.execSync('npm root').toString().trim();
const projectRoot = fs.resolve(npmRoot, "..");
const getProjectRoot = () => projectRoot;
```

0x4007: ChatGPT:

```typescript
import * as path from 'path';
import * as fs from 'fs';

function searchFoundryToml(): string | null {
  /**
   * Recursively searches for a file called "foundry.toml" in parent directories up to the root of the project.
   * Returns the absolute path of the file if found, or null if not found.
   */
  const rootDir = path.parse(process.cwd()).root;
  let currentDir = process.cwd();
  while (true) {
    const foundryTomlPath = path.join(currentDir, 'foundry.toml');
    if (fs.existsSync(foundryTomlPath) && fs.statSync(foundryTomlPath).isFile()) {
      return foundryTomlPath;
    }
    // Reached the root directory of the project, break the loop
    if (currentDir === rootDir) {
      break;
    }
    currentDir = path.dirname(currentDir);
  }
  return null;
}
```

packages/contracts/package.json
d-roak: Since SMTChecker is included in the foundry config file, I can't find how to ignore the SMTChecker when running `forge build` (in order to separate the smt from the compile).

I thought about 'over-engineering' it and run the script to add/remove the SMTChecker config from the `foundry.toml`. But it seems like a bad solution.

Do you have any ideas?

0x4007: I have basically zero formal Foundry experience so it's best if we get input from some of the team. @rndquu @zgorizzo69 @0xcodercrane

rndquu:
I think we shouldn't add smt check directly to the CI build step because the `forge build` time increases greatly (I don't know the exact timing, but smt check has already been running for 60 mins for 8 targets for 217 contracts on my laptop)

What we can do:
1. Create a separate script like `npm run build:smt` which only runs smt check. We can modify `foundry.toml` via js to add `profile.default.model_checker` option (which is not a great solution) or use a plain `solc` with a shell script, whatever fits best.
2. Create a github workflow that runs smt check only on`workflow_dispatch` (not on every PR, because build time is increased greatly)
3. Create a task "Fix SMT errors"

So steps 1 and 2 should be part of the current issue.

packages/contracts/scripts/prebuild/export-smtchecker-contracts.ts
rndquu: I would keep it simple and use a relative import

If the path fails to resolve in the future then CI action for smt (check my last comment) will fail and we will know that smth is wrong

d-roak:
@rndquu 
For step 1, do you prefer that I use `foundry.toml` or a sh script?

In case you prefer the sh script, do you want me to generate the target contracts from the files in the sources or do you want it hardcoded?

rndquu:
> @rndquu For step 1, do you prefer that I use `foundry.toml` or a sh script?
> 
> In case you prefer the sh script, do you want me to generate the target contracts from the files in the sources or do you want it hardcoded?

I think it's easier to use js to edit `foundry.toml` and then run `forge build`, but you're free to use whatever you think works best.

Nothing should be hardcoded, contracts should be searched and dynamically added to the `foundry.toml` file

packages/contracts/scripts/prebuild/export-smtchecker-contracts.ts
0x4007: Okay sounds good.

packages/contracts/package.json
zgorizzo69: what about creating a smt profile https://book.getfoundry.sh/config/?highlight=profile#configuring-with-foundrytoml so i.e 
if `FOUNDRY_PROFILE` is smtchecker then will run smt checker when running forge build otherwise it will not

packages/contracts/foundry.toml
```diff
@@ -22,4 +22,89 @@ bracket_spacing = false
...
+
+[profile.default.model_checker]
+engine = 'chc'
+timeout = 10_000
+targets = [
+    'assert',
+    'constantCondition',
+    'divByZero',
+    'outOfBounds',
+    'overflow',
```
zgorizzo69, line 35: why not `popEmptyArray` and `balance` ?

packages/contracts/scripts/prebuild/export-smtchecker-contracts.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+  });
+
+  return arrayOfFiles;
+};
+
+const main = () => {
+  const buffer = fs.readFileSync(foundry_file);
+  const fileContent = buffer.toString();
+  let contracts = getAllFiles("../../src", []);
```
zgorizzo69, line 26: I don't get why we pass that empty array

zgorizzo69: I think we should use [glob](https://www.npmjs.com/package/glob) instead cos here we are also checking Mock and interfaces we could easily solve that just passing a glob like `src/**/!(Mock*|I*)/*.sol` and it will probably be faster anyway

zgorizzo69 (changes requested):

packages/contracts/package.json
rndquu: > what about creating a smt profile https://book.getfoundry.sh/config/?highlight=profile#configuring-with-foundrytoml so i.e if `FOUNDRY_PROFILE` is smtchecker then will run smt checker when running forge build otherwise it will not

Good point. @jpldcarvalho you could create 2 profiles: default (without smt checker) and the one which enables smt. Then run `forge build` using the smt profile.

rndquu:
Closing it because of inactivity

PR #618, opened 2023-03-31
Fix: Corrupted HTML Structure Error

### LINKED ISSUE #517: Corrupted HTML Structure on Login Animation Component
This unfortunately slipped through because our UbiquiBot was down. The animation from https://github.com/ubiquity/ubiquity-dollar/pull/477 introduced a non matching tag error for the React component. This throws an error in the UI console. 

I spoke with @Steveantor in the dms about it so making this issue as a formality.

Steveantor:
Resolves #517


0x4007:
You're a real chad. I spent longer than I would like to admit trying to figure this one out!

packages/dapp/components/layout/Layout.tsx
```diff
@@ -23,13 +23,25 @@ function ErrorHandler({ error }: { error: Error }) {
...
 export default function Layout({ children }: LayoutProps) {
+  const firstUpdate = useRef(false);
+  const updateFinish = useRef(false);
+
   useEffect(() => {
-    const { ethereum } = window;
-    if (ethereum) {
-      ethereum.on("accountsChanged", window.location.reload);
-      ethereum.on("chainChanged", window.location.reload);
+    //@note Fix: (Illegal Invocation)
```
0x4007, line 30: Can you elaborate?

packages/dapp/components/lib/AppContextProvider.tsx
```diff
@@ -5,7 +5,9 @@ import { UseWeb3Provider } from "./hooks/useWeb3";
 
 import { combineComponents } from "./combineComponents";
 
-const providers = [UseWeb3Provider, ManagedContractsContextProvider, TransactionsContextProvider, BalancesContextProvider];
+// const providers = [UseWeb3Provider, ManagedContractsContextProvider, TransactionsContextProvider, BalancesContextProvider];
+// @note Fix: (Error: missing revert data in call exception since Diamond isn't deployed yet)
```
0x4007, line 9: Should we create an issue for this?

packages/dapp/pages/_app.tsx
```diff
@@ -5,6 +5,7 @@ import "./styles/dark-light-mode-toggle.css";
 
 import { AppProps } from "next/app";
 import Head from "next/head";
+import Script from "next/script"; //@note Fix: (next/script warning)
```
0x4007, line 8: You might not need to prefix `@note` because that doesn't seem to be a standard that I've seen in UI code.

packages/dapp/pages/bonds.tsx
```diff
@@ -1,5 +1,7 @@
 import { FC } from "react";
-import BondsContainer from "../components/bonds/App";
+
+import dynamic from "next/dynamic";
+const BondsContainer = dynamic(() => import("../components/bonds/App"), { ssr: false }); //@note Fix: (Hydration Error)
```
0x4007, line 4: Can you explain to me why this solves the problem? For my education.

Steveantor: well it's slightly complicated but in general, Nextjs renders the page on the server and hydrates on the client
if those versions don't match (due to dynamic rendering such as based on the isConnected/walletAddress)
it throws the hydration error
here's a nice article explaining this in more detail
https://jonmeyers.io/blog/fix-client-server-hydration-error-in-next-js
@pavlovcik

packages/dapp/pages/_app.tsx
Steveantor: I've used this extension that highlights such tags like `@note` `@todo` etc 
https://marketplace.visualstudio.com/items?itemName=alefragnani.Bookmarks
@pavlovcik

packages/dapp/components/lib/AppContextProvider.tsx
Steveantor: yes, I had a discussion in the group about this one @pavlovcik

packages/dapp/components/layout/Layout.tsx
Steveantor: It fixes the Illegal Invocation, likely causing by duplicate event listeners @pavlovcik

PR #620, opened 2023-04-04
Feat: Privatekey or Mnemonic Fallback

### LINKED ISSUE #588: development deploy should either use ENV or constant/mnemonic
if you only have `PRIVATE_KEY` set in env and run `yarn workspace @ubiquity/contracts deploy:development` it will fail cos it takes your private key and use the mnemonic test to get the publickey

Steveantor:
Resolves #588


0x4007:
I followed up again with @zgorizzo69 to check

zgorizzo69 (approved):
good job !

PR #621, opened 2023-04-05
refactor: use local anvil for tests

rndquu:
We don't have any remote tests (that require a forked environment) anymore so we can use a plain anvil instance to:
1. Speed up test runs
2. Eliminate the risk of exceeding RPC rate limits


Steveantor:
We might comment some fields out for future references

packages/contracts/package.json
```diff
@@ -22,7 +22,7 @@
     "tsx": "^3.12.2"
   },
   "scripts": {
-    "test": "tsx ./scripts/runner/runner.ts",
+    "test": "forge test",
```
0x4007: I've noticed issues in the past with conflicting arguments/parameters for shell to pass to yarn -> forge. Are you sure this is stable? Perhaps its better to do: 

```suggestion
    "forge:test": "forge test",
```

0x4007:
> We might comment some fields out for future references

Might be cleaner to tag this commit and delete the code. If we need to restore it we can look for the tag? Seems like a more proper way to maintain and archive the code.

rndquu:
> We might comment some fields out for future references

I don't really like the commented or unused code because it clutters the workspace

Steveantor:
> > We might comment some fields out for future references
> 
> Might be cleaner to tag this commit and delete the code. If we need to restore it we can look for the tag? Seems like a more proper way to maintain and archive the code.

that works too

0x4007:
Unfortunately I am not an expert on best practices for tagging but it can be done retroactively. Can one of you figure out what the best solution is here in what to write as the tag title?

packages/contracts/package.json
rndquu: > Are you sure this is stable? 

Yes, the tests are passing in this PR

> Perhaps its better to do `"forge:test": "forge test"`

We already have `test:slither` and `test:echidna` so perhaps `test:unit` also works

Steveantor:
> Unfortunately I am not an expert on best practices for tagging but it can be done retroactively. Can one of you figure out what the best solution is here in what to write as the tag title?

I've added the tag @pavlovcik

Steveantor (approved):

0x4007 (approved):

zgorizzo69 (approved):
good job !

PR #651, opened 2023-05-20
Fix: call() should be used instead of transfer() to transfer ether

### LINKED ISSUE #645: `call()` should be used instead of `transfer()` to transfer ether.
## Summary
`call()` should be used instead of `transfer()` to transfer ether.

Plase view;
- https://www.theblock.co/post/225364/zksync-unfreeze-millions-stuck

## Vulnerability Detail
The `transfer()` and `send()` functions forward a fixed amount of 2300 gas. Historically, it has often been recommended to use these functions for value transfers to guard against reentrancy attacks. However, the gas cost of EVM instructions may change significantly during hard forks which may break already deployed contract systems that make fixed assumptions about gas costs. For example. EIP 1884 broke several existing smart contracts due to a cost increase of the SLOAD instruction.

## Impact
The use of the deprecated transfer() function for an address will inevitably make the transaction fail when:

- The claimer smart contract does not implement a payable function.
- The claimer smart contract does implement a payable fallback which uses more than 2300 gas unit.
- The claimer smart contract implements a payable fallback function that needs less than 2300 gas units but is called through proxy, raising the call's gas usage above 2300.
- Additionally, using higher than 2300 gas might be mandatory for some multisig wallets.

## Code Snippet

- [UbiquiStickSale.sol#L122](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/ubiquistick/UbiquiStickSale.sol#L122)
- [UbiquiStickSale.sol#L128](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/ubiquistick/UbiquiStickSale.sol#L128)
- [LibCollectableDust.sol#L65](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/libraries/LibCollectableDust.sol#L65)

rotcivegaf:
Resolves #645

### DIFF
--- a/packages/contracts/src/dollar/libraries/LibCollectableDust.sol
+++ b/packages/contracts/src/dollar/libraries/LibCollectableDust.sol
@@ -62,7 +62,8 @@ library LibCollectableDust {
             "collectable-dust::token-is-part-of-the-protocol"
         );
         if (_token == ETH_ADDRESS) {
-            payable(_to).transfer(_amount);
+            (bool result, ) = _to.call{value: _amount}("");
+            require(result, "Failed to send Ether");
         } else {
             IERC20(_token).safeTransfer(_to, _amount);
         }
--- a/packages/contracts/src/ubiquistick/UbiquiStickSale.sol
+++ b/packages/contracts/src/ubiquistick/UbiquiStickSale.sol
@@ -119,12 +119,14 @@ contract UbiquiStickSale is Ownable, ReentrancyGuard {
         // Calculate any excess/unspent funds and transfer it back to the buyer
         if (msg.value > paid) {
             uint256 unspent = msg.value - paid;
-            payable(msg.sender).transfer(unspent);
+            (bool result, ) = msg.sender.call{value: unspent}("");
+            require(result, "Failed to send Ether");
             emit Payback(msg.sender, unspent);
         }
     }
 
     function withdraw() public nonReentrant onlyOwner {
-        payable(fundsAddress).transfer(address(this).balance);
+        (bool result, ) = fundsAddress.call{value: address(this).balance}("");
+        require(result, "Failed to send Ether");
     }
 }


0x4007:
### GPT-4 Review:

In the context of Ethereum smart contracts, the changes made in this pull request replace the `transfer` function with a `call` function to transfer Ether. 

In the file `LibCollectableDust.sol`, the change is:

```diff
- payable(_to).transfer(_amount);  
+ (bool result, ) = _to.call{value: _amount}("");  
+ require(result, "Failed to send Ether");
```

In the file `UbiquiStickSale.sol`, similar changes are made in two places:

```diff
- payable(msg.sender).transfer(unspent);
+ (bool result, ) = msg.sender.call{value: unspent}("");
+ require(result, "Failed to send Ether");

- payable(fundsAddress).transfer(address(this).balance);
+ (bool result, ) = fundsAddress.call{value: address(this).balance}("");
+ require(result, "Failed to send Ether");
```

The reason behind this change is likely due to a known potential vulnerability associated with the `transfer` function. The `transfer` function in Solidity only forwards 2300 gas, which might not be enough if the receiving contract needs to do anything complex in its fallback function. This limit is imposed to discourage potential reentrancy attacks, but it might also cause legitimate transactions to fail.

On the other hand, the `call` function forwards all remaining gas, allowing the receiver's fallback function to perform more complex operations. However, it could potentially expose the contract to reentrancy attacks if not used properly.

But in this case, the `call` function is used safely because the `require` statement is used to ensure that the call was successful. Additionally, these changes are made in non-reentrant functions or in places where state changes have already happened, which further mitigates the risk of reentrancy attacks.

In conclusion, the changes appear to be secure and should enhance the functionality of the contract by allowing more complex operations in the receiver's fallback function.

0x4007 (approved):

zgorizzo69 (approved):
Lgtm 👏

PR #655, opened 2023-05-24
Resolves #647 - Use EIP-1967

### LINKED ISSUE #647: `LibAppStorage` is not using `EIP-1967`
### Description:

Instead of using `EIP-1967` as in all others facets, the `AppStorage` is setted on slot `0`; [LibAppStorage.sol#L47](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/libraries/LibAppStorage.sol#L47)

### Code:
- [LibAppStorage.sol#L38-L44](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/libraries/LibAppStorage.sol#L38-L44)

### Recommendation:
Use `LibAppStorage.appStorage()` instead of `store` to accessung the store, also its recommend to use differents store slots for each variable, and instead of changing directly it should have setters and getters for each variable.

```diff
diff --git a/packages/contracts/src/dollar/libraries/LibAppStorage.sol b/packages/contracts/src/dollar/libraries/LibAppStorage.sol
index 148ee493..7b52c905 100644
--- a/packages/contracts/src/dollar/libraries/LibAppStorage.sol
+++ b/packages/contracts/src/dollar/libraries/LibAppStorage.sol
@@ -36,16 +36,18 @@ struct AppStorage {
 }
 
 library LibAppStorage {
+    bytes32 constant private APP_STORAGE_STORAGE_SLOT = 
+        bytes32(uint256(keccak256("ubiquity.contracts.curve.storage")) - 1);
+    
     function appStorage() internal pure returns (AppStorage storage ds) {
+        bytes32 position = APP_STORAGE_STORAGE_SLOT;
         assembly {
-            ds.slot := 0
+            ds.slot := position
         }
     }
 }
 
 contract Modifiers {
-    AppStorage internal store;
-
     /**
      * @dev Prevents a contract from calling itself, directly or indirectly.
      * Calling a `nonReentrant` function from another `nonReentrant`
```

PhantomCracker:
Resolves #647

### DIFF
--- a/packages/contracts/src/dollar/libraries/LibAppStorage.sol
+++ b/packages/contracts/src/dollar/libraries/LibAppStorage.sol
@@ -91,11 +91,43 @@ contract Modifiers {
         // https://eips.ethereum.org/EIPS/eip-2200)
         store.reentrancyStatus = store.NOT_ENTERED;
     }
-            
+
     modifier onlyOwner() {
         LibDiamond.enforceIsContractOwner();
         _;
     }
+    
+    modifier onlyAdmin() {
+        require(
+            LibAccessControl.hasRole(DEFAULT_ADMIN_ROLE, msg.sender),
+            "Manager: Caller is not admin"
+        );
+        _;
+    }
+
+    modifier onlyMinter() {
+        require(
+            LibAccessControl.hasRole(GOVERNANCE_TOKEN_MINTER_ROLE, msg.sender),
+            "Governance token: not minter"
+        );
+        _;
+    }
+
+    modifier onlyBurner() {
+        require(
+            LibAccessControl.hasRole(GOVERNANCE_TOKEN_BURNER_ROLE, msg.sender),
+            "Governance token: not burner"
+        );
+        _;
+    }
+
+    /**
+     * @dev Modifier to make a function callable only when the contract is not paused.
+     *
+     * Requirements:
+     *
+     * - The contract must not be paused.
+     */
 
     modifier onlyCreditNFTManager() {
         require(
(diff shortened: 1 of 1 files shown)


packages/contracts/src/dollar/libraries/LibAppStorage.sol
```diff
@@ -117,8 +128,17 @@ contract Modifiers {
      *
      * - The contract must not be paused.
      */
+
+    modifier onlyCreditNFTManager() {
```
0x4007, line 132:
```suggestion
    modifier onlyCreditNftManager() {
```

Please see the [coding style guide](https://github.com/ubiquity/ubiquity-dollar/blob/development/.github/CONTRIBUTING.md) for more information.

0x4007 (commented):
Looks mostly good to me but will need @zgorizzo69 review to officially approve and merge. I just realized that you didn't pass CI. You unfortunately might need to force-push and clobber your commit messages to adhere to the conventional commits specification.

zgorizzo69:
> Looks mostly good to me but will need @zgorizzo69 review to officially approve and merge. I just realized that you didn't pass CI. You unfortunately might need to force-push and clobber your commit messages to adhere to the conventional commits specification.

IMHO  We can merge even though the commit is not following the convention but we can't if it the build/test workflow doesn't pass

zgorizzo69:
also I don't think we should use EIP-1967 for app storage...

0x4007:
Okay so in conclusion we should close and not merge but pay the bounty out because it has been completed according to the spec. Please confirm @zgorizzo69

PhantomCracker:
@pavlovcik @zgorizzo69 any news here? Because I would like to jump on another bounty.

rndquu:
We should not have paid the bounty because the build is failing so the task is not ready even if we considered it as "not needed"

0x4007:
> We should not have paid the bounty because the build is failing so the task is not ready even if we considered it as "not needed"

Valid point thanks for noting this.

PR #676, opened 2023-06-08
QA for cloudflare deployment

cuddle-dev:
Resolves #


0x4007:
You should be careful to open draft pull requests by default, and then convert it to a "real" pull request if you need review because we are now using the codeowners feature.

cuddle-dev:
> You should be careful to open draft pull requests by default, and then convert it to a "real" pull request if you need review because we are now using the codeowners feature.

This is just for a QA no need to merge it so once it was tested, I closed it

PR #688, opened 2023-06-20
feat: enforce kebab-case filenames

0x4007:


packages/dapp/fixtures/contracts-addresses/.keep
0x4007, line 1: Not sure if this can be deleted.

packages/dapp/public/mQBuVsUlmV.txt
0x4007, line 1: Not sure if this can be deleted.

Issue #684, opened 2023-06-17
Gas optimization, reentrancy, unused variable and unused code after a return statement was called


malik672:
1.Many of the functions can be reduced to view or pure , this will probably save you 20x the original gas fees used
2. There is a potential renetrancy in using openzepplin onERC1155received when you make external calls to another address
3. there are a lot of unused variable in the contract this will 100% increase gas fees 
4. a lot of variables are being shadowed, this will 100% result in the contract being exploited 
5. there's a function that updates a state variable after the return statement, this won't work because everything after a return variable won't work
6. You can probably reduce gas fees for this contract by 76% with the introduction of assembly and bitwise operation
7. There might be more, but this is what i've seen

0x4007:
Thanks for filing this issue.

2. should be validated before we make a bounty for it. 
3. linters should handle this
4. linters handle this. 
5. which function?
6. this is very unlikely to be the decision we make because code maintainability is more important. However, it could be interesting to optimize the popular functions.

rndquu:
> 1.Many of the functions can be reduced to view or pure , this will probably save you 20x the original gas fees used 2. There is a potential renetrancy in using openzepplin onERC1155received when you make external calls to another address 3. there are a lot of unused variable in the contract this will 100% increase gas fees 4. a lot of variables are being shadowed, this will 100% result in the contract being exploited 5. there's a function that updates a state variable after the return statement, this won't work because everything after a return variable won't work 6. You can probably reduce gas fees for this contract by 76% with the introduction of assembly and bitwise operation 7. There might be more, but this is what i've seen

hi, thank you for creating an issue

1. we will leave it for the "gas optimisation" task
2. you are free to create a PR with POC and fix, I think @pavlovcik will set a bounty accordingly
3. we will leave it for the "gas optimisation" task
4. we will leave it for the "slither analysis" task
5. which one?
6. we will leave it for the "gas optimisation" task

malik672:
@pavlovcik @rndquu
```
   /// @notice returns the percentage of deviation from the peg multiplied by amount
    //          when Ubiquity Dollar is <1$
    function _getPercentDeviationFromUnderPeg(
        uint256 amount
    ) internal returns (uint256) {
        uint256 curPrice = _getTWAPPrice();
        if (curPrice >= 1 ether) {
            return 0;
        }

        bytes16 res = _one.sub(curPrice.fromUInt()).mul(amount.fromUInt());
[...]

PR #691, opened 2023-06-20
fix: remove unreachable return

rndquu:
This PR removes an unreachable return mentioned [here](https://github.com/ubiquity/ubiquity-dollar/issues/684#issuecomment-1597166172)

### DIFF
--- a/packages/contracts/src/dollar/libraries/LibCurveDollarIncentive.sol
+++ b/packages/contracts/src/dollar/libraries/LibCurveDollarIncentive.sol
@@ -159,7 +159,6 @@ library LibCurveDollarIncentive {
         bytes16 res = _one.sub(curPrice.fromUInt()).mul(amount.fromUInt());
         // returns (1- TWAP_Price) * amount.
         return res.div(_one).toUInt();
-        LibTWAPOracle.update();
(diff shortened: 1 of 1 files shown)


0x4007 (approved):
Makes sense to me

pavlovcik:
Shouldn't we run `LibTWAPOracle.update();` before the `return` ?

0x4007 (commented):
Shouldn't we run `LibTWAPOracle.update();` before the `return` ? Updating the oracle seems like a good idea.

rndquu:
> Shouldn't we run `LibTWAPOracle.update();` before the `return` ? Updating the oracle seems like a good idea.

It is already updated in all of the functions where `_getPercentDeviationFromUnderPeg()` is used ([one](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/libraries/LibCurveDollarIncentive.sol#L112) and [two](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/libraries/LibCurveDollarIncentive.sol#L135))

0x4007 (approved):

PR #693, opened 2023-06-20
refactor: kebab-case file names

0x4007:
Something I've wanted for a long time. 

Git isn't normally case sensitive so relying on casing for file names is a bad idea. It also introduces discrepancies around how to structure casing (some folks want to capitalize acronyms, and many do not agree on the max length of these acronyms, leading to several situations with "Nft" and "NFT" in file names.) 

Kebab case makes the most sense for file naming in the context of git, regardless of what apparently most Solidity projects use (camel case.)


pavlovcik:
It builds but I don't understand why test coverage is complaining.

0x4007:
> [Deployment: Tue Jun 20 2023 21:16:57 (UTC)](https://fc1735b2.ubiquity-dollar-5vd.pages.dev)

CSS styles are not looking good

packages/contracts/scripts/deploy/manager.ts
0x4007, line 1: I might have broken the deployments here I think I need to change back the contract names but I am not too sure how to test deploys.

0x4007:
Can I get some extra eyes on the deployment related stuff? The build and the UI works so from my perspective it seems functional.

This is urgent to merge in because this will force bounty hunters to restart their active tasks.

cuddle-dev:
> Can I get some extra eyes on the deployment related stuff? The build and the UI works so from my perspective it seems functional.
> 
> This is urgent to merge in because this will force bounty hunters to restart their active tasks.

If we have several active PRs, hunters would not complain about that due to the conflicts?

0x4007:
> > Can I get some extra eyes on the deployment related stuff? The build and the UI works so from my perspective it seems functional.
> 
> > 
> 
> > This is urgent to merge in because this will force bounty hunters to restart their active tasks.
> 
> 
> 
> If we have several active PRs, hunters would not complain about that due to the conflicts?

Honestly just gonna need to push this through asap and have the hunters append their changes again to the latest. 

I plan to merge this in within 24 hours of being open or else it's not gonna happen (again) 

This is something that's been attempted a couple of times already but we never merged in due to various reasons. One of them being the reason you mentioned here. 

Please help me verify the solidity side is okay with deployments etc.

packages/contracts/scripts/deploy/manager.ts
rndquu: This is an old deployment script so I guess it can be deleted

0x4007:
`packages/contracts/scripts/deploy/dollar/solidityScripting/*`

I had to remove the numerical file name prefixes because the build process was failing when typeschain wanted to turn these into variables. Identifiers can not begin with numbers in typescript.

rndquu:
There is an official solidity style guide which [explicitly states](https://docs.soliditylang.org/en/v0.8.16/style-guide.html#contract-and-library-names) that:
```
Contracts and libraries should be named using the CapWords style.
Contract and library names should also match their filenames.
If the contract name is `Congress` and the library name is `Owned`, then their associated filenames should be `Congress.sol` and `Owned.sol`.
```

If kebab naming is so critical then we should take an iterative approach to it:
1. Convert config files to kebab case
2. Convert ts/js files to kebab case
3. etc..
4. **Leave solidity files in the CapWords style**

No need in a mega PR which updates the whole project.

To sum up:
1. We shouldn't update solidity files to kebab case because we end up with such **inconsistent** code:
```
import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
import {CreditNft} from "../../dollar/core/credit-nft.sol";
```
2. This PR introduces us more bugs than profit because everything is updated (yml configs, ts files, solidity files, etc...)

I would close this PR and refactor to kebab case step by step without touching solidity files.

@0xcodercrane @zgorizzo69 help

.github/workflows/coverage-check.yml
```diff
@@ -106,13 +106,18 @@ jobs:
...
           exit 1
 
-    - name: Coveralls GitHub Action
-      uses: coverallsapp/github-action@v1
-      with:
-        github-token: ${{ secrets.COVERALLS_REPO_TOKEN }}
+      - name: Coveralls GitHub Action
+        uses: coverallsapp/github-action@v1
+        with:
+          github-token: ${{ secrets.COVERALLS_REPO_TOKEN }}
```
0x4007, line 118: I'm not sure how this worked before. It said there was a syntax error.

.github/workflows/enforce-kebab-case.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Check For Non Kebab-Cased File Names
+        run: |
+          non_compliant_files=()
+          while read -r file; do
+            basefile=$(basename "$file")
+            if echo "$file" | grep -q -E "^\.\/\..*\/|^\.\/[^\/]*$|^\.\/node_modules|^\.\/\.git|^\.\/dist|^\.\/build|^\.\/vendor|^\.\/\.next|\.sql$|\.md$|LICENSE|echidna\.config\.yml|\.env\.example|slither\.config\.json|\.npmrc|global\.d\.ts|_app\.tsx|next-env\.d\.ts|next\.config\.js"; then
```
0x4007, line 17: We can ignore `.sol` files.

0x4007:
> Leave solidity files in the CapWords style

I think that it could make sense. Conceptually my understanding is that each `.sol` file is supposed to represent only a single smart contract, which is capitalized when its a variable (identifier.) 

So the file is, in a way, representing the contents of the identifier which I think could make sense to leave with the same exact file name as the identifier.

0x4007:
No Solidity file changes: https://github.com/ubiquity/ubiquity-dollar/pull/694

PR #711, opened 2023-07-01
fix: check foundry builds

molecula451:
We'll keep an eye out on this was, the way they release (foundry), it broke the commit hash.

This PR only updates foundry's commit hash

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -27,7 +27,6 @@ jobs:
         uses: foundry-rs/foundry-toolchain@v1
         with:
           version: nightly-31d6498c79af595577f200fc2136b31f43885397
-          
 
       - name: Yarn Install
         run: yarn install --mode=skip-build && yarn allow-scripts
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Setup Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly-ed5eb9796aefe8cccc8d56676aa90ff68d7edfd4
+          version: nightly-31d6498c79af595577f200fc2136b31f43885397
 
       - name: Yarn Install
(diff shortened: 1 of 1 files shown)


rndquu:
This PR basically fixes the "Build & Test" workflow, good to merge

rndquu (approved):

rndquu:
@molecula451 next time pls follow conventional commits naming

nopeless:
@rndquu Their workflow checks for release hashes, and `31d6498c79af595577f200fc2136b31f43885397` is not a release hash

molecula451:
it was, a few days ago, then it changes and gets lost

nopeless:
@molecula451 that shouldn't happen unless they force delete refs. do you think thats the case?

molecula451:
> @molecula451 that shouldn't happen unless they force delete refs. do you think thats the case?

that why the issue #715  was opened, to take a look at that until give with the root cause, because we rely on CI, so not trivial, but we need latest foundry, and not any older commit

nopeless:
> > @molecula451 that shouldn't happen unless they force delete refs. do you think thats the case?
> 
> that why the issue #715 was opened, to take a look at that until give with the root cause, because we rely on CI, so not trivial, but we need latest foundry, and not any older commit

We may have to ask the foundry maintainers about this. Deleting ref tags is not proper of them. Do you think it might be better to use a stable foundry commit for now?

molecula451:
> We may have to ask the foundry maintainers about this. Deleting ref tags is not proper of them. Do you think it might be better to use a stable foundry commit for now?


This PR set a stable commit, it got lost again, it's been twice, so we'll check that out on, to have a stable commit hash

nopeless:
> > We may have to ask the foundry maintainers about this. Deleting ref tags is not proper of them. Do you think it might be better to use a stable foundry commit for now?
> 
> 
> This PR set a stable commit, it got lost again, it's been twice, so we'll check that out on, to have a stable commit hash

By stable, I meant an old commit (>1 month) that will less likely be deleted by any kind of git operations (e.g. rebase)

PR #710, opened 2023-07-01
Update UI - fix some bugs

molecula451:
Resolves #695 

This PR is for UI, altho there are minimal changes in contracts folders too

### What this great PR does

- This PR gets  rid of anything bonds related, according to **_pavlovcik_** and confirmed by **_rndquu_**
- This PR fix a bug related to invalid-key (this is `WalletConnect ID` relate, pass a wallet connect ID to secrets)
-  This PR introduces and new file [local-data.ts](https://github.com/ubiquity/ubiquity-dollar/blob/fbe4164f7218dee4901d931e8330a0b9376e9143/packages/dapp/components/utils/local-data.ts) and new method `fetchData()` to DEBUG local contract instances, deployed to anvil and on chain ID 31337
-  This PR adds `dotenv` to handle `DEBUG` var on .`env`
-  This PR purges and unused packaged named `hardhat` obviously because the team does not want anything with `hardhat`
-  This PR DOES NOT clash production, because it needs `DEBUG=true` to perform, so it's only locally
-  Cleans and update 404 page

### Room for Subtaskes

- This PR DOES NOT, but needs refactor, `use-managed-manager.tsx`, it's doing nothing and the code refactor and debug

PR #714, opened 2023-07-02
Fix: slither errors

### LINKED ISSUE #472: UbiquiStick slither checks
Right now we have [slither disabled](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/slither.config.json#L8) for this [folder](https://github.com/ubiquity/ubiquity-dollar/tree/development/packages/contracts/src/ubiquistick).

What should be done:
1. Remove `src/ubiquistick` from `filter_paths` in the [slither config file](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/slither.config.json)
2. Fix slither errors in the `src/ubiquistick` folder

> Ideally it should be included in slither checks but I figured out it requires a lot of changes. Fixing slither issue would take lots of time rather than re-implementing them properly. so I made it excluded. In the near future, we should include again.

_Originally posted by @0xcodercrane in https://github.com/ubiquity/ubiquity-dollar/pull/417#discussion_r1067061553_

ghost:
Resolves #472


ghost:
@rndquu there is this slither error about 
image

but the thing is the code is already checking the return value but slither seems to not detect it what should we do about it?

rndquu:
@AnakinSkywalkeer pls merge the latest development branch to fix the "Build & Test" workflow

rndquu:
> @rndquu there is this slither error about image
> 
> but the thing is the code is already checking the return value but slither seems to not detect it what should we do about it?

Yes, the code is fine, we should not (in this case) check the 2nd param with return data, only the result. There seems to be a [bug](https://github.com/crytic/slither/issues/2008) in slither so you are free to [disable](https://github.com/crytic/slither/wiki/Usage#triage-mode) this warning. We should not exclude the detector entirely because it has the `medium` severity.

ghost:
@rndquu there is this error about Arbitrary from in transferFrom they suppose us to use the msg.sender in the transferfrom but we don't really wanna do that in this case 
image
the from should be the treasury should we ignore it?

ghost:
> @AnakinSkywalkeer pls merge the latest development branch to fix the "Build & Test" workflow

My branch is already upto date with development branch seems like smth is wrong with Foundry

rndquu:
> @rndquu there is this error about Arbitrary from in transferFrom they suppose us to use the msg.sender in the transferfrom but we don't really wanna do that in this case image the from should be the treasury should we ignore it?

So `treasury` approved the `SimpleBond` contract to spend ERC20 tokens. It means that any user can call some method in the `SimpleBond` contract where `safeTransferFrom(treasury, to, amount)` is used and drain the whole treasury. But the claim amount is restricted by the [staked amount](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/ubiquistick/SimpleBond.sol#L178) so I don't see ways for an end user to drain the whole treasury.

So this way we can set slither to ignore this [line](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/ubiquistick/SimpleBond.sol#L186). Besides, as far as I understand, the `SimpleBond` contract is not used anywhere and somewhat deprecated. This needs @zgorizzo69 's confirmation.

ghost:
@rndquu there is this last error its about the usage of block.timestamp to create the randomness 
image

this is the function which is using it hm should we use any other way to create the random like chainlink VRF etc may be 🤔

rndquu:
> @rndquu there is this last error its about the usage of block.timestamp to create the randomness image
> 
> this is the function which is using it hm should we use any other way to create the random like chainlink VRF etc may be 🤔

Yes, miner can control most of the params to some extent. The `UbiquiStick` contract is on hold at the moment so I think we can omit the error for now.

.github/workflows/build-and-test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Setup Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly-31d6498c79af595577f200fc2136b31f43885397
+          version: nightly-9aefa433dfa28dd74b83c83f7265c91f7506f4bc
```
rndquu: pls merge the latest `development` branch which has the same foundry version

.github/workflows/slither-analysis.yml
```diff
@@ -1,11 +1,7 @@
 name: Slither Analysis
 
-on:
-  push:
-    branches:
-      - development
-      - main
-  pull_request:
+on: [push, pull_request]
```
rndquu: Why was it updated?

packages/contracts/src/dollar/core/UbiquityCreditToken.sol
```diff
@@ -5,8 +5,9 @@ import {ManagerFacet} from "../facets/ManagerFacet.sol";
 import {ERC20Ubiquity} from "./ERC20Ubiquity.sol";
 import {IERC20Ubiquity} from "../../dollar/interfaces/IERC20Ubiquity.sol";
 import "../libraries/Constants.sol";
+import {IUAR} from "../../ubiquistick/interfaces/IUAR.sol";
 
-contract UbiquityCreditToken is ERC20Ubiquity {
+contract UbiquityCreditToken is ERC20Ubiquity, IUAR {
```
rndquu: `IUAR` is somewhat deprecated, pls remove the inheritance

packages/contracts/src/ubiquistick/UbiquiStickSale.sol
```diff
@@ -101,18 +101,20 @@ contract UbiquiStickSale is Ownable, ReentrancyGuard {
 
         if (remainingTokenCount < count) {
             count = remainingTokenCount;
+            paid = remainingTokenCount;
```
rndquu: `paid` should be equal to the funds paid, not to the number of remaining tokens

packages/contracts/src/ubiquistick/UbiquiStickSale.sol
```diff
@@ -101,18 +101,20 @@ contract UbiquiStickSale is Ownable, ReentrancyGuard {
...
         }
         if (msg.value < count * price) {
+            paid = (msg.value * price) / price;
```
rndquu: What is the purpose of multiplying and dividing to the same `price` variable ?

packages/contracts/src/ubiquistick/UbiquiStickSale.sol
```diff
@@ -101,18 +101,20 @@ contract UbiquiStickSale is Ownable, ReentrancyGuard {
...
             count = msg.value / price;
         }
         if (MAXIMUM_PER_TX < count) {
+            paid = MAXIMUM_PER_TX;
```
rndquu: `paid` should be equal to the amount of funds paid, not to the number of max allowed tokens to buy

rndquu (changes requested):

packages/contracts/src/ubiquistick/UbiquiStickSale.sol
ghost: > What is the purpose of multiplying and dividing to the same `price` variable ?

image
the count calculated using the msg.value/ price and then there was paid variable in the bottom which was multiplying the count with price but the slither wants us to multiply the price first and then divide the thing but putting paid = count/ price and gonna break stuff so just to make sure that paid gets the right value i thought to change the paid variable in the conditional functions

rndquu: Can't we refactor to:
```
paid = msg.value;
```

?

ghost: > Can't we refactor to:
> 
> ```
> 
> paid = msg.value;
> 
> ```
> 
> 
> 
> ?

Sure my bad I didn't look 😅😅😅we can

packages/contracts/src/ubiquistick/UbiquiStickSale.sol
```diff
@@ -93,26 +92,28 @@ contract UbiquiStickSale is Ownable, ReentrancyGuard {
...
             count > 0,
             "Not Whitelisted For The Sale Or Insufficient Allowance"
         );
-
         if (remainingTokenCount < count) {
             count = remainingTokenCount;
+            paid = remainingTokenCount * price;
         }
         if (msg.value < count * price) {
+            paid = msg.value * price;
```
rndquu: Shouldn't it be `paid = count * price` inserted after this [line](https://github.com/ubiquity/ubiquity-dollar/pull/714/files#diff-6bb51bf2f1f9f376618af61a61a3adb284a5b42d097fc1c9e91973e0a3b0b810R107)?

rndquu (changes requested):

packages/contracts/src/ubiquistick/UbiquiStickSale.sol
ghost: > Shouldn't it be `paid = count * price` inserted after this [line](https://github.com/ubiquity/ubiquity-dollar/pull/714/files#diff-6bb51bf2f1f9f376618af61a61a3adb284a5b42d097fc1c9e91973e0a3b0b810R107)?

sorry my bad i was supposed to remove that * price too its gonna be just msg.value

rndquu (approved):
The "Compare Test Coverage" workflow is [failing](https://github.com/ubiquity/ubiquity-dollar/actions/runs/5469123835/jobs/9957655980?pr=714) but the changes are purely cosmetic so we can put up with it

packages/contracts/src/ubiquistick/SimpleBond.sol
```diff
@@ -135,10 +135,13 @@ contract SimpleBond is ISimpleBond, Ownable, Pausable {
         // @dev must set token allowance for this smart contract previously
         IERC20(token).safeTransferFrom(msg.sender, address(this), amount);
 
-        Bond memory bnd;
-        bnd.token = token;
-        bnd.amount = amount;
-        bnd.block = block.number;
+        Bond memory bnd = Bond({
```
molecula451: nice change, can you rename the state variable to something clearer, readable, that does not clash with the `Bond struct`

ghost: > nice change, can you rename the state variable to something clearer, readable, that does not clash with the `Bond struct`

sure

molecula451:
@AnakinSkywalkeer sync branch with `development` as there are new changes

packages/contracts/src/ubiquistick/SimpleBond.sol
```diff
@@ -183,6 +186,7 @@ contract SimpleBond is ISimpleBond, Ownable, Pausable {
 
             assert(bnd.claimed <= bnd.rewards);
```
molecula451: update the state var here, and in 180, too https://github.com/ubiquity/ubiquity-dollar/pull/714/files#diff-39a1832a63be8f745df33562f0b2f0a49b3e69356f50e357e05de5a1e47d5feeR180

molecula451:
> @rndquu there is this error about Arbitrary from in transferFrom they suppose us to use the msg.sender in the transferfrom but we don't really wanna do that in this case image the from should be the treasury should we ignore it?

no, we don't, in either case add a `require()` to make sure, it's `treasury` address

packages/contracts/src/dollar/core/UbiquityCreditToken.sol
molecula451: remove the import https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L10 @AnakinSkywalkeer

molecula451:
since I agree with @rndquu in that if we're not going to use this contract in the future, then there is not reason to keep adding cosmetic changes, then add the last changes, and I am approving the rest of the task seems good @AnakinSkywalkeer

.github/workflows/slither-analysis.yml
molecula451: @AnakinSkywalkeer may you put some wording here?

rndquu: this is already fixed

packages/contracts/src/dollar/core/UbiquityCreditToken.sol
rndquu: > remove the import
> 
> https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L10
> 
> @AnakinSkywalkeer

this `IUAR` is used [here](https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L188)

packages/contracts/src/ubiquistick/SimpleBond.sol
rndquu: You mean renaming `bnd` to `bondState` ?

molecula451: to enhance readability, for further uses, if we're going to keep the contract

rndquu:
> > @rndquu there is this error about Arbitrary from in transferFrom they suppose us to use the msg.sender in the transferfrom but we don't really wanna do that in this case image the from should be the treasury should we ignore it?
> 
> no, we don't, in either case add a `require()` to make sure, it's `treasury` address

The [treasury](https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L109) address is not controlled by an external user and can only be set by the contracts' owner. What's the point of `require(treasury == treasury)` ?

molecula451:
> The [treasury](https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L109) address is not controlled by an external user and can only be set by the contracts' owner. What's the point of `require(treasury == treasury)` ?

it might be of help to have the check, to prevent calling from external actor, on `safeTransferFrom`, rather have a `REVERT`, i'm judging by initial comment on the "trasurey drain". to ensure reliability in probably need to be battle test in different vectors, tho

rndquu:
Could you suggest the updated code?

molecula451:
> Could you suggest the updated code?

We know `safeTransferFrom`, apply it's "safetiness", but im in favor of checks with `REVERT`, with claiming things
suggestion needs testing: 

```solidity
require(treasury == msg.sender, "not treasury");
IERC20(tokenRewards).safeTransferFrom(
    treasury,
    msg.sender,
    claimAmount
);
```

also I'm seeing there's room for `nonReentrant` on this call

rndquu:
> > Could you suggest the updated code?
> 
> We know `safeTransferFrom`, apply it's "safetiness", but im in favor of checks with `REVERT`, with claiming things suggestion needs testing:
> 
> ```solidity
> require(treasury == msg.sender, "not treasury");
> IERC20(tokenRewards).safeTransferFrom(
>     treasury,
>     msg.sender,
>     claimAmount
> );
> ```
> 
> also I'm seeing there's room for `nonReentrant` on this call

But it's going to always revert because the [claimBond()](https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L177) is meant to be called by EOAs, not by the treasury address

molecula451:
> > > Could you suggest the updated code?
> > 
> > 
> > We know `safeTransferFrom`, apply it's "safetiness", but im in favor of checks with `REVERT`, with claiming things suggestion needs testing:
> > ```solidity
> > require(treasury == msg.sender, "not treasury");
> > IERC20(tokenRewards).safeTransferFrom(
> >     treasury,
> >     msg.sender,
> >     claimAmount
> > );
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > also I'm seeing there's room for `nonReentrant` on this call
> 
> But it's going to always revert because the [claimBond()](https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L177) is meant to be called by EOAs, not by the treasury address

Yeah i judged by the initially comment on drain up, to have the contributor look for the best check on `REVERT`, else if we're going to keep the contract, then we line by line apply safety measure on a new issue

rndquu:
> > > > Could you suggest the updated code?
> > > 
> > > 
> > > We know `safeTransferFrom`, apply it's "safetiness", but im in favor of checks with `REVERT`, with claiming things suggestion needs testing:
> > > ```solidity
> > > require(treasury == msg.sender, "not treasury");
> > > IERC20(tokenRewards).safeTransferFrom(
> > >     treasury,
> > >     msg.sender,
> > >     claimAmount
> > > );
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > also I'm seeing there's room for `nonReentrant` on this call
> > 
> > 
> > But it's going to always revert because the [claimBond()](https://github.com/ubiquity/ubiquity-dollar/blob/21f0514176801209f71e85ed8e6cbfef4f36e385/packages/contracts/src/ubiquistick/SimpleBond.sol#L177) is meant to be called by EOAs, not by the treasury address
> 
> Yeah i judged by the initially comment on drain up, to have the contributor look for the best check on `REVERT`, else if we're going to keep the contract, then we line by line apply safety measure on a new issue

To sum up, what else do we expect from @AnakinSkywalkeer to fix? Only this [issue](https://github.com/ubiquity/ubiquity-dollar/pull/714#discussion_r1254796466)?

molecula451:
> To sum up, what else do we expect from @AnakinSkywalkeer to fix? Only this [issue](https://github.com/ubiquity/ubiquity-dollar/pull/714#discussion_r1254796466)?

Let's have him push the changes and good to close

packages/contracts/src/ubiquistick/SimpleBond.sol
molecula451: good reability == good maintanibility find akward this vars so short this way

rndquu:
@AnakinSkywalkeer pls fix issue in this [comment](https://github.com/ubiquity/ubiquity-dollar/pull/714#discussion_r1254796466)

molecula451:
> @AnakinSkywalkeer pls fix issue in this [comment](https://github.com/ubiquity/ubiquity-dollar/pull/714#discussion_r1254796466)

rndquu, have us confirm that we will keep this contract, and I will personally put on a night review line by line and testing for further issues

rndquu:
> > @AnakinSkywalkeer pls fix issue in this [comment](https://github.com/ubiquity/ubiquity-dollar/pull/714#discussion_r1254796466)
> 
> rndquu, have us confirm that we will keep this contract, and I will personally put on a night review line by line and testing for further issues

`SimpleBond` along with `UbiquiStick` are on hold right now so there is no need to go to great length for them

This one https://github.com/ubiquity/ubiquity-dollar/pull/710 is 10x more important

molecula451 (approved):

rndquu:
The "Compare Test Coverage" workflow is failing but code changes are purely cosmetic so a slight decrease is negligible

PR #716, opened 2023-07-03
fix: replace broken release hash

### LINKED ISSUE #715: Watch for Foundry's commit hash releases
relates #711

Foundry has a distinctive way of updating new releasing it looks like some commit hashes get lost when this happens.

We are relying on the commit hash for our actions CI, so we need to make sure everything's in harmony with the hash

nopeless:
Resolves #711

I don't expect pay

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Setup Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly-31d6498c79af595577f200fc2136b31f43885397
+          version: nightly-9aefa433dfa28dd74b83c83f7265c91f7506f4bc
 
       - name: Yarn Install
         run: yarn install --mode=skip-build && yarn allow-scripts


nopeless:
Rationale: https://github.com/foundry-rs/foundry/releases includes 9aefa433dfa28dd74b83c83f7265c91f7506f4bc exists while 31d6498c79af595577f200fc2136b31f43885397 is not present

molecula451:
can you avoid closing #715 ?

nopeless:
@molecula451 no problem. Do you want me to write a short script that fetches latest valid release?

rndquu:
@nopeless thank you for this PR, let's merge it. If the `nightly-9aefa433dfa28dd74b83c83f7265c91f7506f4bc` is also deleted then we should probably checkout to some older (last month?) version. 

I guess https://github.com/foundry-rs/foundry-toolchain uses https://github.com/foundry-rs/foundry/tags for the `version` workflow input and the previous tag `nightly-31d6498c79af595577f200fc2136b31f43885397` was simply deleted

PR #727, opened 2023-07-09
improve: follow linter conventions

molecula451:
Contract improvement, by following linter conventions

solidity `fallback()` functions makes **_low-level_** checks whether or not `receive()` exists on the contract, for gas amusements and other stuff

this was even purged in a original diamond reference https://github.com/mudgen/diamond-2/commit/40aeb5043f904c820454a0b17afdc3430e31f7d5

I also don't see uses to keep it as `fallback()` is already **payable**, but! we may follow the conventional linter for dev improvement and overall contracts measurement


rndquu:
@molecula451 this PR is marked as draft, is it ready for review?

0x4007:
Oh sorry it was not clear to me that it is a draft when I requested your review on mobile.

molecula451:
> @molecula451 this PR is marked as draft, is it ready for review?

not yet. I'm opening for review after opened PRs turns around

molecula451:
The linter! sometimes, but not always, throws potential unnecessary advices, but this one seems interesting, also, I don't see any withdraw-able function to use out of this one, so if ether gets into it, it'll get stuck

0x4007:
> so if ether gets into it, it'll get stuck

That seems like an issue!

molecula451:
> > so if ether gets into it, it'll get stuck
> 
> That seems like an issue!

Absolutely!

Options we have:

1. Completely purge it as unnecesary
2. add `REVERT` to it. People sending ether to the contract will get revert

```solidity 
receive() external payable {
   //no ether allowed
   revert()
 }
```
 
3. Add an admin withdraw-able function
4. Keep as it but follow the linter convention as `fallback()` checks if exists

rndquu:
@molecula451 

1. This is the original [Diamond implementation](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/Diamond.sol). This is the PR's [Diamond implementation](https://github.com/ubiquity/ubiquity-dollar/blob/96f42402602d9ae19f25e242ad60dfe6db07ed30/packages/contracts/src/dollar/Diamond.sol). The only difference is the order of functions so basically those 2 contracts are equal. What issue this PR solves?
2. What linter command produces an error regarding the `receive()` method?

molecula451:
@rndquu 
This PR's only makes sure `fallbacks()` notice `receive()` exists to handle gas amusement as on a low-level end `fallbacks()` checks for it, and also it looks like it triggers some options https://github.com/ubiquity/ubiquity-dollar/pull/727#issuecomment-1630526116,

2. It's part of the Dev env, not really integrated into the project

rndquu:
I would simply remove the `receive()` method because we don't use it

molecula451:
> I would simply remove the `receive()` method because we don't use it

Yeah we are in fact, in harmony, purging

molecula451:
@rndquu done in https://github.com/ubiquity/ubiquity-dollar/pull/727/commits/7aa35ca0bf7fc43e333c6cd698cf44ece36af21c

rndquu (approved):

0x4007:
I don't have any new information or perspective to bring to this pull request.

PR #734, opened 2023-07-17
feat: create tsx for cow zap transaction

### LINKED ISSUE #610: Add ZAP to collateral contract
There is the [UbiquityPoolFacet](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol) contract which allows users to:
- transfer any ERC20 tokens to the contract in exchange for `uAD` 
- redeem `uAD` tokens from the contract in exchange for the collateral ERC20 tokens

We should have the ability to swap any ERC20 token to `LUSD` under the hood when user [deposits](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol#L17) collateral in order to mint `uAD`. 

Notice that:
- swaps should be made via the [CoW Swap](https://swap.cow.fi/).
- there might be edge cases on [mint](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol#L17) or [redeem](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol#L33) when there is not enough liquidity in the pools

Original description:
- we want user to be able to **ZAP** In any token to be swapped to `LUSD` to retrieve `uAD` as described in #609

hashedMae:
Resolves #610


hashedMae:
Working on the tests. It's a super simple contract for facilitating the zap. Just need to get the Ubiquity Pool setup to do it.

packages/dapp/components/ubiquity-pool/cow-zap.jsx
0x4007, line 1: We only use `.tsx` so this should be renamed and turned into a TypeScript file.

0x4007 (changes requested):

0x4007:
> Just need to get the Ubiquity Pool setup to do it.

Can you elaborate on what you mean by this?

packages/dapp/components/ubiquity-pool/cow-zap.jsx
hashedMae: Can do

hashedMae:
> > Just need to get the Ubiquity Pool setup to do it.
> 
> Can you elaborate on what you mean by this?

The contracts are what is called by CoW Protocol to interact with the UbiquityPoolFacet contract. The setup from the previous tests worked so it really didn't take much.

molecula451:
i am drafting this PR as it looks it needs more work to be done, @hashedMae request review when you really think you accomplish the task

molecula451:
@rndquu there might be some progress here! but I think this PR, in it's actual stage doesn't solve the issue #610

1. Should we close?
2. Should we give an opportunity to the worker with new thoughts, etc? and no dismiss the progress

I think we should close in favor of some thoughts from the hunter (means give some time to the hunter to provide some thoughts)

rndquu:
> @rndquu there might be some progress here! but I think this PR, in it's actual stage doesn't solve the issue #610
> 
> 1. Should we close?
> 2. Should we give an opportunity to the worker with new thoughts, etc? and no dismiss the progress
> 
> I think we should close in favor of some thoughts from the hunter (means give some time to the hunter to provide some thoughts)

I don't understand why you set this PR as ready for review from the draft state

> I think we should close in favor of some thoughts from the hunter

It's actually handled automatically by the bot so let the bot does his job

0x4007:
> It's actually handled automatically by the bot so let the bot does his job

It's supposed to but it missed this one due to the pagination issue. It unassigned mae two weeks ago. I'll close for it being stale.

PR #749, opened 2023-08-09
ci: run slither check on .sol files update

rndquu:
Slither workflow runs for ~15min. We don't really need it to run on each PR or on push to the `development` branch. We should run it only when `.sol` files are updated. 

This PR modifies slither workflow to run only when solidity files are updated.

### DIFF
--- a/.github/workflows/slither-analysis.yml
+++ b/.github/workflows/slither-analysis.yml
@@ -4,7 +4,11 @@ on:
     branches:
       - development
       - main
+    paths:
+      - '**.sol'
   pull_request:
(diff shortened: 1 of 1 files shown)


0x4007:
Does it work on every solidity file no matter how deeply nested it is inside directories? It's hard for me to tell from my phone.

rndquu:
> Does it work on every solidity file no matter how deeply nested it is inside directories? It's hard for me to tell from my phone.

yes

0x4007 (approved):

PR #753, opened 2023-08-10
refactor: remove metapool migration scripts

rndquu:
https://crv.to is back online so we don't need Curve MetaPool migration scripts anymore


0x4007:
On one hand, makes sense to delete what is not needed. On the other hand, seems like a waste to throw out this work. I want to do some more research to verify that there are no benefits for migrating to v3 before accepting this pull request. 

https://github.com/curvefi/curve-docs/blob/3fcb5f25d4acec0fb445f1f480a8b8ed3b589dca/docs/exchange-lp-tokens.rst#curve-token-v3

- It looks like v3 has gas optimizations and two new methods to mitigate race conditions: `increaseAllowance` and `decreaseAllowance`
- > For Curve LP Tokens V1 and V2, non-zero to non-zero approvals are prohibited. Instead, after every non-zero approval, the allowance for the spender must be reset to 0.

The only reason I hesitate to migrate is in case something goes wrong with the migration. This is not my area of expertise.

rndquu:
> https://github.com/curvefi/curve-docs/blob/3fcb5f25d4acec0fb445f1f480a8b8ed3b589dca/docs/exchange-lp-tokens.rst#curve-token-v3

You're referring to v3 Curve LP token while Curve Metapool is a different contract

v2 Metapool: https://etherscan.io/address/0x20955CB69Ae1515962177D164dfC9522feef567E#code
v3 Metapool: https://etherscan.io/address/0x9558b18f021FC3cBa1c9B777603829A42244818b#code

There are only minor updates between the 2 versions. Besides [v3](https://etherscan.io/address/0x9558b18f021FC3cBa1c9B777603829A42244818b#code) (which we were intended to migrate to) is compiled with vyper 0.2.15 where reentrancy lock doesn't work due to the compiler bug.

In my opinion we don't get many benefits from this migration hence the code is removed.

0x4007:
> where reentrancy lock doesn't work due to the compiler bug.

Oh really so we should totally avoid v3 then!

PR #755, opened 2023-08-10
Revamp regex to fix CodeQL scanning security (& performance) fix

### LINKED ISSUE #746: Fix code scanning alert - Inefficient regular expression
Tracking issue for:
- [x] https://github.com/ubiquity/ubiquity-dollar/security/code-scanning/11

molecula451:
Resolves #746

as per github security and the Code QL action, it checks the regex that it's prone to degrade performance. This should fix it, but we will still take a look on the security tab and make sure it's gone

This update, does not interfere, with `deploy:development`

### DIFF
--- a/packages/contracts/scripts/deploy/manager/diamond-in-all.ts
+++ b/packages/contracts/scripts/deploy/manager/diamond-in-all.ts
@@ -93,7 +93,7 @@ const create = async (args: ForgeArguments): Promise<{ result: DeploymentResult 
   }
   if (stdout) {
     //test regex
-    const regex = /[^{}]*(?:R[^{}]*)*}/g;
+    const regex = /[^{}]*+(?:R[^{}]*+)*+}/g;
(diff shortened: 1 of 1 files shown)


molecula451:
i'm seeing exploitable patterns if only we were to run this a client-side, this regex is running under a dev scripting enviroment, so we instead take this regex as a "performance degradation" on our repo and not really and exploitable pattern as mention on github security, since it's not really "client-side"

molecula451:
mergin this as this file is prone to be change as per #756

0x4007:
> mergin this as this file is prone to be change as per #756 

I don't understand but you shouldn't merge without review

molecula451:
> > mergin this as this file is prone to be change as per #756
> 
> I don't understand but you shouldn't merge without review

Clear, thing, it's probably this file it's gonna be purge at #756 or totally revamped as it looks like a very old, unused script

PR #761, opened 2023-08-11
ci: generate docs only on sol files update

rndquu:
Right now solidity natspec docs are generated on every push to the `development` branch which is really necessary when `.sol` files are updated

So this PR makes sure that docs are regenerated only when `.sol` files are updated


molecula451 (approved):
absolutely

PR #768, opened 2023-08-21
feature: deep fuzzing on newly modified Solidity contracts and pull requests

### LINKED ISSUE #515: Deep Fuzzing on Newly Modified Solidity Files
Random thought I had but since I [wrote that action](https://github.com/ubiquity/bounty-bot/blob/c1c6c99336f34ac5e94efaed49c9f218fb7a2d76/.github/workflows/push.yml#L36-L43) that will: 

1. read the bounty-bot repo and 
3. build if any source code changes occur, and then 
4. commit the built file to the repo 

It inspired me to think of what other compute we can delegate to the cloud to make the dev experience better. 

What if we do something similar with deep fuzzing, but only on modified Solidity files? I understand that deep fuzzing takes a lot of time, so we could run it for the entire codebase one time, and then only run it on modified Solidity files moving forward? I vaguely recall a discussion related to how deep fuzzing is super slow.

transmissions11 has a Foundry template that does a deep fuzz on everything when merged to main, which served as another point of inspiration for me.

```yml
      - name: Run tests
        run: forge test
        env:
          # Only fuzz intensely if we're running this action on a push to main or for a PR going into main:
          FOUNDRY_PROFILE: ${{ (github.ref == 'refs/heads/main' || github.base_ref == 'main') && 'intense' }}
```
https://github.com/transmissions11/foundry-template/blob/0bf8e53589f2e8f72e41baf672c25aa35ba57295/.github/workflows/tests.yml#L28-L32

gitcoindev:
Resolves #515

### DIFF
--- a/.github/workflows/deep-fuzz.yml
+++ b/.github/workflows/deep-fuzz.yml
@@ -1,0 +1,59 @@
+name: Deep Fuzz
+on:
+  push:
+  pull_request:
+
+jobs:
+  deep-fuzz:
+    name: Deep Fuzz
+    runs-on: ubuntu-22.04
+    env:
+      FOUNDRY_PROFILE: intense
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "18.14.1"
+
+      - name: Setup Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly-d369d2486f85576eec4ca41d277391dfdae21ba7
+
+      - name: Get changed Solidity contracts
+        id: changed-files
+        uses: tj-actions/changed-files@v37
+        with:
+          files: packages/contracts/src/**
+
+      - name: Forge install
+        working-directory: packages/contracts
+        run: forge install
+
+      - name: Deep Fuzz Solidity Contracts
+        working-directory: packages/contracts
+        run: |
+          if ${{ github.event_name == 'pull_request' }}; then
+            forge test
+          fi
+          if ${{ github.event_name == 'push' }}; then
+            if [ -z "${{ steps.changed-files.outputs.all_changed_files }}" ]; then
+              echo "No Solidity contracts were modified, skipping deep fuzzing..."
+              exit 0
+            fi
+            touch fuzz.txt
+            for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+              echo "${file} was changed"
+              f=$(basename -- $file)
+              forge tree | grep -E "^test|^src|${f}" | grep -B1 ${f} | grep ^test | cut -d' ' -f1 >> fuzz.txt
+            done
+            echo "Deep fuzzing tests:"
+            cat fuzz.txt | sort | uniq -u
+            cat fuzz.txt | sort | uniq -u | xargs -I{} forge test --match-path {}
+          fi
--- a/.github/workflows/deep-fuzz.yml
+++ b/.github/workflows/deep-fuzz.yml
@@ -12,17 +12,10 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       FOUNDRY_PROFILE: intense
-    permissions:
-      contents: read
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-
-      - name: Setup Node
-        uses: actions/setup-node@v3
-        with:
-          node-version: "18.14.1"
 
       - name: Setup Foundry
         uses: foundry-rs/foundry-toolchain@v1
--- a/.github/workflows/deep-fuzz.yml
+++ b/.github/workflows/deep-fuzz.yml
@@ -1,7 +1,10 @@
 name: Deep Fuzz
 on:
   push:
-  pull_request:
+    branches:
+      - development
+    paths:
+      - '**.sol'
 
 jobs:
   deep-fuzz:
@@ -26,34 +29,10 @@ jobs:
         with:
           version: nightly-d369d2486f85576eec4ca41d277391dfdae21ba7
 
-      - name: Get changed Solidity contracts
-        id: changed-files
-        uses: tj-actions/changed-files@v37
-        with:
-          files: packages/contracts/src/**
-
       - name: Forge install
         working-directory: packages/contracts
         run: forge install
 
       - name: Deep Fuzz Solidity Contracts
         working-directory: packages/contracts
-        run: |
-          if ${{ github.event_name == 'pull_request' }}; then
-            forge test
-          fi
-          if ${{ github.event_name == 'push' }}; then
-            if [ -z "${{ steps.changed-files.outputs.all_changed_files }}" ]; then
-              echo "No Solidity contracts were modified, skipping deep fuzzing..."
-              exit 0
-            fi
-            touch fuzz.txt
-            for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
-              echo "${file} was changed"
-              f=$(basename -- $file)
-              forge tree | grep -E "^test|^src|${f}" | grep -B1 ${f} | grep ^test | cut -d' ' -f1 >> fuzz.txt
-            done
-            echo "Deep fuzzing tests:"
-            cat fuzz.txt | sort | uniq -u
-            cat fuzz.txt | sort | uniq -u | xargs -I{} forge test --match-path {}
-          fi
+        run: forge test
--- a/packages/contracts/foundry.toml
+++ b/packages/contracts/foundry.toml
@@ -41,4 +41,4 @@ targets = [
 src = 'src/dollar'
 
 [profile.intense.fuzz]
-runs = 10000
+runs = 100000000
--- a/packages/contracts/foundry.toml
+++ b/packages/contracts/foundry.toml
@@ -39,3 +39,6 @@ targets = [
 
 [profile.docs]
 src = 'src/dollar'
+
+[profile.intense.fuzz]
+runs = 10000
--- a/packages/contracts/foundry.toml
+++ b/packages/contracts/foundry.toml
@@ -41,4 +41,4 @@ targets = [
 src = 'src/dollar'
 
 [profile.intense.fuzz]
-runs = 100000000
+runs = 100000


rndquu:
Regarding the [issue](https://github.com/ubiquity/ubiquity-dollar/issues/515):
1. We don't have proper fuzzing tests right now so we're getting ahead of ourselves with this PR
2. There is no sense in running fuzzing tests only on modified solidity files because all of the contracts are interconnected so in the end we'd better to deep fuzz the whole protocol 

Regarding the PR:
1. If we had proper deep fuzzing tests then it would take for them a decent amount of time to finish. I don't think that we should run this "deep fuzz" workflow on `pull_request` or `push` events. Perhaps it is better to run the workflow on `workflow_dispatch` or on push to the `development` branch.
2. Isn't it better to use the workflow syntax ([example](https://github.com/ubiquity/ubiquity-dollar/blob/development/.github/workflows/slither-analysis.yml#L8)) to check for modified files instead of a shell script?

gitcoindev:
Hi @rndquu thank you for starting the review and your valuable comments. 

> 2. Isn't it better to use the workflow syntax ([example](https://github.com/ubiquity/ubiquity-dollar/blob/development/.github/workflows/slither-analysis.yml#L8)) to check for modified files instead of a shell script?

I used the recommended workflow implementation for listing changed files for https://github.com/tj-actions/changed-files , see https://github.com/tj-actions/changed-files#using-local-git-history . Since deep fuzz will be run on the whole protocol I will use the Slither's example that you provided.

Apart from that I will incorporate your feedback and submit updates so that the pull request can be accepted. 
I would also like to keep @pavlovcik in the loop so that he is aware of the changes.

gitcoindev:
Hi @rndquu , @pavlovcik I incorporated the suggested changes https://github.com/ubiquity/ubiquity-dollar/pull/768/commits/cfd27ede08ca45b992b3c8b3c0b5bbd886bac624 , my QA is available at https://github.com/gitcoindev/ubiquity-dollar/actions/runs/5935230402/job/16093344234 . The pull request is ready for another round of the review.

0x4007:
> There is no sense in running fuzzing tests only on modified solidity files because all of the contracts are interconnected so in the end we'd better to deep fuzz the whole protocol

Warning: I did not read the fuzz documentation on Foundry, but from what I understand about the concepts, here are my beliefs: the concept of fuzzing is testing as many types of inputs as possible. We only need to deep fuzz the entire protocol once, and then afterwards only deep fuzz the changed contracts. It does not matter if the contracts are connected or not with the rest of the protocol, because each contract is fuzz tested in isolation anyways. It will only add more time to fuzz test the entire protocol every time a Solidity file is changed. 

>  see https://github.com/tj-actions/changed-files#using-local-git-history

You should be using the GitHub API version, not relying on local .git history (we are not running locally) which checks for changed files on a pull request: https://github.com/tj-actions/changed-files#using-githubs-api

gitcoindev:
> > There is no sense in running fuzzing tests only on modified solidity files because all of the contracts are interconnected so in the end we'd better to deep fuzz the whole protocol
> 
> Warning: I did not read the fuzz documentation on Foundry, but from what I understand about the concepts, here are my beliefs: the concept of fuzzing is testing as many types of inputs as possible. We only need to deep fuzz the entire protocol once, and then afterwards only deep fuzz the changed contracts. It does not matter if the contracts are connected or not with the rest of the protocol, because each contract is fuzz tested in isolation anyways. It will only add more time to fuzz test the entire protocol every time a Solidity file is changed.
> 
> > see https://github.com/tj-actions/changed-files#using-local-git-history
> 
> You should be using the GitHub API version, not relying on local .git history (we are not running locally) which checks for changed files on a pull request: https://github.com/tj-actions/changed-files#using-githubs-api

Hi @pavlovcik @rndquu , I executed a few experiments during implementation. Currently 35 out of 267 total tests in packages/contracts are fuzz tests, with default 256 runs for each test case. The deep fuzzing takes about 2-3 minutes for 10000 runs for the whole protocol, so currently this is still manageable.

For my original approach with QA here https://github.com/gitcoindev/ubiquity-dollar/issues/2 : I had to find out which tests imported changed contracts and execute 10000 runs on those. When the test code base grows significantly this approach can still be used.  

I also already reworked the pull request to use GitHub API version and verified it triggers only on push to development branch when *.sol files change: see https://github.com/gitcoindev/ubiquity-dollar/actions/runs/5935230402

packages/contracts/foundry.toml
```diff
@@ -39,3 +39,6 @@ targets = [
 
 [profile.docs]
 src = 'src/dollar'
+
+[profile.intense.fuzz]
+runs = 10000
```
0x4007: If this only takes 2-3 minutes, why not crank this up to something excessive? Perhaps something that will get us close to the GitHub Actions max runner runtime of 6 hours?

~1,440,000 runs? We can round it down to 1 million runs.

gitcoindev: Hi @pavlovcik 1 000 000 took ~20 minutes on my quite powerful laptop. This is due to fuzzer logic and increasing corpus which stores coverage.

But it may not make sense to set this to such large value anyway, see https://consensys.io/diligence/blog/2023/04/benchmarking-smart-contract-fuzzers/  , at some point fuzzer corpus saturates and it does not find more issues. Imho we could try to set to 100 000 , which currently takes ~7 minutes . Please let me know and I will update the value to the one you suggest.

0x4007: In the article they said they ran each fuzzing "campaign" for eight hours so based on that we should aim for the full six hours on the GitHub runner. 

It also mentions caching so perhaps we can preserve the cache by uploading the artifact, and then downloading it on the next run? This can be a separate bounty. 

Lastly I think it's very interesting how different fuzzers have such different results. We should finish implementing Echidna and look into Harvey. @rndquu

gitcoindev: @rndquu could you please also add your 2 cents here - does six hours makes sense to enable now? I would be rather for setting to 100 000 and adding more fuzz tests or enable invariant testing or both and then after some time extend to 6 hours when more data is gathered. I somehow feel that without having more data pushing this hamster on a wheel for six hours would just consume resources without visible benefits.

0x4007: > would just consume resources 

GitHub Actions is a free resource which is why we should take advantage of it.

gitcoindev: @pavlovcik ok, sure, shall I set the value to 100 000 000 runs ?

gitcoindev: done

.github/workflows/deep-fuzz.yml
```diff
@@ -0,0 +1,38 @@
...
+    paths:
+      - '**.sol'
+
+jobs:
+  deep-fuzz:
+    name: Deep Fuzz
+    runs-on: ubuntu-22.04
+    env:
+      FOUNDRY_PROFILE: intense
+    permissions:
```
rndquu: Unnecessary permission, pls remove

.github/workflows/deep-fuzz.yml
```diff
@@ -0,0 +1,38 @@
...
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup Node
```
rndquu: Unnecessary step, pls remove

packages/contracts/foundry.toml
```diff
@@ -39,3 +39,6 @@ targets = [
 
 [profile.docs]
 src = 'src/dollar'
+
+[profile.intense.fuzz]
+runs = 100000000
```
rndquu: The "Deep Fuzz" workflow with `1 000 000` runs takes 72 minutes to finish for a github runner. I guess setting it to `100 000 000` would simply exceed the 6 hour runner limit.

Right now our fuzzing tests are unstructured and in some cases don't really make sense. For example [here](https://github.com/ubiquity/ubiquity-dollar/blob/8fa34df8b37903f2e812cfacd2f1ba438faae095/packages/contracts/test/dollar/core/StakingShare.t.sol#L140) is the fuzz test for minting to zero address. The next [one](https://github.com/ubiquity/ubiquity-dollar/blob/8fa34df8b37903f2e812cfacd2f1ba438faae095/packages/contracts/test/dollar/core/StakingShare.t.sol#L150) tests for minting with not enough permissions. Fuzzing tests should be run against protocol invariants, not against permissions or input validation.

I agree that with the current fuzzing tests implementation it doesn't make much sense in running the "Deep Fuzz" workflow for 6 hours. I think we should set the `runs` param to `100 000` and increase this number when we have decent fuzzing tests.

rndquu (changes requested):

packages/contracts/foundry.toml
0x4007: I don't understand the conservative approach with essentially a free resource

rndquu: Just because a resource is free doesn't mean that we should exploit it. There should be a reasonable approach.

Right now we don't have proper fuzzing tests that are run against protocol invariants. Indeed some of the fuzzing tests are useful but the other part is not. So we'd better make our fuzzing tests better and only then utilize github to the full 6 hours.

.github/workflows/deep-fuzz.yml
gitcoindev: done

.github/workflows/deep-fuzz.yml
gitcoindev: done

packages/contracts/foundry.toml
gitcoindev: I addressed @rndquu 's comments, and the number of runs is the only thing remaining to merge this pull request.

I did a 100 000 000 fuzz run: https://github.com/gitcoindev/ubiquity-dollar/actions/runs/5972597368/job/16203392834 which was canceled after 6 hours. No violations were currently found by the fuzzer. 

Unfortunately, time based fuzzing is not supported yet in Forge: https://github.com/foundry-rs/foundry/issues/4517

@pavlovcik , @rndquu please agree on a value of runs to set for now, and I will update according to your wish.

rndquu: Since `1 000 000` runs take 72 minutes to finish I guess `10 000 000` runs will also exceed the 6 hour CI limit. So set it either to `1 000 000` either to `100 000`.

gitcoindev: All right, seems reasonable. I will set 100 000 and when amount of tests increases the value can be always modified. This will allow for fast feedback loop. Perhaps until more tests are developed the time based fuzzing will be available.

gitcoindev: done

rndquu (approved):

gitcoindev:
Thank you for approvals and merging!

PR #790, opened 2023-09-19
Bonding curve

### LINKED ISSUE #611: Algorithmic Market Operations for Collateral
There is the [UbiquityPoolFacet](https://github.com/ubiquity/ubiquity-dollar/blob/development/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol) where users can:
1. Deposit collateral ERC20 tokens in exchange for Dollar tokens
2. Redem Dollar tokens in exchange for the collaterral

We should put the provided collateral to work and earn some yield.

What should be done:
1. Take any AMO strategy contract (from frax finance or some other well audited protocol)
2. Create an ERC-4626 contract over that strategy so that we could easily add new AMO strategies to the ubiquity protocol
3. Add thorough natspec comments for all of the contracts, methods and mechanics
4. Add unit testing
 
Good article: https://mirror.xyz/jackchong.eth/iB_teRKgBaKm4OTKFmjf8hFAM55C1_i0Z2kf-KkYz2I

_Original description:_
we want an AMO for collateral gathered thanks to  https://github.com/ubiquity/ubiquity-dollar/issues/609
Algorithmic Market Operations  module is an autonomous contract(s) that enacts arbitrary  monetary policy so long as it does not change the uAD price off its peg. This means that AMO controllers can perform open market operations algorithmically (as in the name), but they cannot arbitrarily mint uAD out of thin air and break the peg. The aims is to invests idle collateral into various DeFi vaults/protocols

FibrinLab:
Resolves #611


rndquu:
Closing because of inactivity

PR #808, opened 2023-10-06
Refactor/diamond test naming

rndquu:
This PR refactors [DiamondTestSetup](https://github.com/rndquu/ubiquity-dollar/blob/8a4fc742ea380d1f8a8ebc825ebabfb448a3ab3d/packages/contracts/test/diamond/DiamondTestSetup.sol) so that all variables started with a lower case letter. The original [DiamondTestSetup](https://github.com/ubiquity/ubiquity-dollar/blob/d41cc0dbb6cb70ff8edd8c9b3487a66db56f132f/packages/contracts/test/diamond/DiamondTestSetup.sol) has [variables](https://github.com/ubiquity/ubiquity-dollar/blob/d41cc0dbb6cb70ff8edd8c9b3487a66db56f132f/packages/contracts/test/diamond/DiamondTestSetup.sol#L64) which start with an upper case letter so it is not clear whether it is an interface or a variable.

### DIFF
--- a/packages/contracts/test/diamond/DiamondTest.t.sol
+++ b/packages/contracts/test/diamond/DiamondTest.t.sol
@@ -17,9 +17,8 @@ contract TestDiamond is DiamondTestSetup {
 
     function testFacetsHaveCorrectSelectors() public {
         for (uint256 i = 0; i < facetAddressList.length; i++) {
-            bytes4[] memory fromLoupeFacet = ILoupe.facetFunctionSelectors(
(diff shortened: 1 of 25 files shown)


gitcoindev:
+1 for this PR. I was also confused before as it looked as interface was called directly, eg. `ILoupe.facetFunctionSelectors()`

molecula451 (approved):
This PR enhances reading & follows project convention with implementation

PR #811, opened 2023-10-07
linter: update eslint flag

### LINKED ISSUE #810: Forbid specific non-null assertion
You're using "!" Everywhere which should have been strictly forbidden as part of our lint settings.

https://github.com/ubiquity/ubiquity-dollar/pull/809#pullrequestreview-1663031693
https://github.com/ubiquity/ubiquity-dollar/pull/809#discussion_r1349496374

molecula451:
Resolves #810

QA:
null

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -6,7 +6,6 @@
   "enableFiletypes": ["solidity"],
 
   "words": [
-    "!",
     "!UbiquityDollar",
     "!UbiquityCredit",
     "!UbiquityGovernance",
--- a/cspell.json
+++ b/cspell.json
@@ -6,6 +6,7 @@
   "enableFiletypes": ["solidity"],
 
   "words": [
+    "!",
     "!UbiquityDollar",
     "!UbiquityCredit",
     "!UbiquityGovernance",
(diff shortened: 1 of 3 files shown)


rndquu:
I don't understand the purpose of this [code change](https://github.com/ubiquity/ubiquity-dollar/pull/811/files#diff-77519f787fe7e051a14c588f101d61e4d8f0d3b01bf8e16cb65d3838243d4e68R9). I tried both with and without the "!" symbol in the `cspell.json` file and it doesn't affect the build.

We already have the `non-null assertion` rule enabled hence the warnings are shown. What we should do is to throw an error if there are some warnings produced by eslint so that the CI build failed with error. This [line](https://github.com/ubiquity/ubiquity-dollar/blob/c119128ef92a5b042f5b136364eda9f537600f80/package.json#L51) could be updated to `eslint --fix --max-warnings=0` ([more info](https://stackoverflow.com/questions/51346302/eslint-throw-errors-instead-of-warnings))

molecula451:
lmao, well the thing is i rarely use `!` in the code, and always fix my code to eslint errors i thought that was why the warning pop by updating the `cspell.json` will have to recheck that a bit

so basically the contributor ignored the warning, we should have updated to a fail/error instead

molecula451:
i think `--max-warnings=0` is approachable so we keep it simple and fix errors when eslint pop em up

rndquu (approved):

PR #847, opened 2023-12-09
chore: Update missing natspec

molecula451:
This is only a chore PR but i've also made sure that not other library is left without natspec


gitcoindev (approved):
Looks safe

molecula451:
@rndquu looks busy with the needed front end on time, let me know for further concerns

PR #853, opened 2023-12-21
chore: Add .solhint.json to enforce kebab case exception list

gitcoindev:
Adds `.solhint.json` file to kebab case exception list.


molecula451 (approved):
thanks for the backup gitcoindev

PR #637, opened 2023-05-17
635 remove unnecessary files from coverage

eugenioclrc:
Resolves #635

In forge currently is impossible to ignore files, but there is a wayaround to tackle this issue.
Please view https://github.com/foundry-rs/foundry/issues/2988 based on this comment https://github.com/foundry-rs/foundry/issues/2988#issuecomment-1505771312 i have setup a `coverage.sh` script to ignore tests, mocks and scriptint solidity files.

PR #864, opened 2024-01-04
chore: deprecate coverage script

### LINKED ISSUE #862: Fix `yarn coverage`
There is the https://github.com/ubiquity/ubiquity-dollar/blob/f99f7e7995a83657b54d7a4b61fe1722e70d1971/packages/contracts/coverage.sh script which produces the following output:
Screenshot 2024-01-02 at 14 32 25

That output is not accurate because coverage for some of the contracts is not accurate.

If you run coverage with `forge coverage` you will see that `LibUbiquityPool` has 100% line coverage while https://github.com/ubiquity/ubiquity-dollar/blob/f99f7e7995a83657b54d7a4b61fe1722e70d1971/packages/contracts/coverage.sh outputs 16.8%.

What should be done:
1. Either fix the [`yarn coverage`](https://github.com/ubiquity/ubiquity-dollar/blob/f99f7e7995a83657b54d7a4b61fe1722e70d1971/packages/contracts/coverage.sh) script either delete it if it's not used anywhere.
2. Make sure that `yarn coverage` outputs "real" coverage values (you can simply call `forge coverage` under the hood)

molecula451:
Resolves #862 
This PR deprecates the coverage.sh script that resides in the contract folder, rndqnuu doesn't seem to get the actual results and also we do have forge coverage that output different results to.

So if this PR gets in it will be a deprecation in favor of forge coverage
If the PR doesn't get in the coverage.sh will still resides, altho there is a [workflow ](https://github.com/ubiquity/ubiquity-dollar/blob/development/.github/workflows/coverage-check.yml)that does it and the issue would be closed as not planned

# coverage.sh results on latest development branch
Screenshot from 2024-01-03 21-37-53

# forge coverage results on latest development branch
Screenshot from 2024-01-03 21-45-08


molecula451:
looks like rndquu introduced an update in favor of forge coverage

https://github.com/ubiquity/ubiquity-dollar/commit/0ba7e3e1b08bdd57f5eafdd7d10329b9329ed94f

gitcoindev:
It will be interesting to find out why the coverage value was wrong. If I am not mistaken it was detected during recent audit.

I found out that the original solution was introduced in https://github.com/ubiquity/ubiquity-dollar/pull/637 and was implemented based on the workaround in a still open foundry issue: https://github.com/foundry-rs/foundry/issues/2988#issuecomment-1298602707

molecula451:
to clarify things, the first coverage.sh (which is rndquu mentions that he gets wrong) output it's the output that you get after run it in a clean clone development branch, not the forwared repo for the audit etc.., the second too it's clean branch here dollar/development

rndquu:
> looks like rndquu introduced an update in favor of forge coverage
> 
> [0ba7e3e](https://github.com/ubiquity/ubiquity-dollar/commit/0ba7e3e1b08bdd57f5eafdd7d10329b9329ed94f)

We urgently needed a working `yarn coverage` with relevant results for the audit start that is why I simply updated `yarn coverage` to use `forge coverage` under the hood.

If `coverage.sh` is not used anywhere I think it's ok to remove it.

rndquu (approved):

gitcoindev:
Indeed, GitHub actions do not use it.

gitcoindev (approved):

Issue #876, opened 2024-01-18
ci: diamond storage check exceeds API rate limit


molecula451:
There is the "Check For Diamond Storage Changes" [workflow](https://github.com/ubiquity/ubiquity-dollar/blob/development/.github/workflows/diamond-storage-check.yml) which checks that storage collisions on contracts upgrades are not introduced. 

Sometimes it throws the "API rate limit exceeded" error which should be fixed.

What should be done:
1. Fix the API rate limit error
2. Enable (i.e. revert this [PR](https://github.com/ubiquity/ubiquity-dollar/pull/878)) storage checks

https://github.com/ubiquity/ubiquity-dollar/actions/runs/7576286107/job/20634758710?pr=875

refer

![Screenshot from 2024-01-18 17-42-28](https://github.com/ubiquity/ubiquity-dollar/assets/41552663/9f02528d-488c-457c-9f4c-6eb7ae7acc3a)

molecula451:
@pavlovcik @rndquu @gitcoindev

gitcoindev:
@molecula451 @pavlovcik @rndquu 

I see a few ways to tackle this: 

1) Throw money at the problem and increase API rate limit to 15000 (may workaround the issue only short term).
2) Wait with next request until after the time specified by the x-ratelimit-reset header, the workflow would have to check this before running and wait in a loop, or we could have another workflow which would collect those errors and try to re-run. See https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit

But, the best would be to collect statistics and figure out which action eats up API rate the most and try to optimize it.

rndquu:
> But, the best would be to collect statistics and figure out which action eats up API rate the most and try to optimize it.

Or run the diamond storage check workflow only for modified contracts and not for all of them

gitcoindev:
> > But, the best would be to collect statistics and figure out which action eats up API rate the most and try to optimize it.
> 
> Or run the diamond storage check workflow only for modified contracts and not for all of them

Yes, let's open a ticket for that, this is very likely the root cause.

molecula451:
> Or run the diamond storage check workflow only for modified contracts and not for all of them

This one sounds like it will take time, for deployment, testing and we are in the middle of the judging, i think we need a fix asap

rndquu:
> > Or run the diamond storage check workflow only for modified contracts and not for all of them
> 
[...]

PR #878, opened 2024-01-19
chore: temporarily disable storage changes workflow for pull requests

gitcoindev:
Temporarily disable for pull requests until limiting rate API issue is resolved.

Reference: #876

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,9 +1,9 @@
 name: CodeQL Analysis
 
-on:
-  push:
-    branches:
-      - development
+on: workflow_dispatch
+  # push:
+  #  branches:
+  #    - development
   # pull_request:  ## temporarily disable for pull requests until https://github.com/ubiquity/ubiquity-dollar/issues/876 is resolved.
 
 jobs:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,8 +1,10 @@
 name: CodeQL Analysis
 
 on:
-  pull_request:
   push:
+    branches:
+      - development
+  # pull_request:  ## temporarily disable for pull requests until https://github.com/ubiquity/ubiquity-dollar/issues/876 is resolved.
 
 jobs:
   code-ql-analysis:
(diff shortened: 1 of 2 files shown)


molecula451 (approved):
but let's get to the root of the cause @gitcoindev, as this is one of the most sophisticated implementation we have that checks for diamond storage @rndquu

gitcoindev:
> but let's get to the root of the cause @gitcoindev, as this is one of the most sophisticated implementation we have that checks for diamond storage

CodeQL triggered the rate limit API error, perhaps it is the root cause or combination of both: CodeQL analysis and storage changes. I disabled this one as well.

rndquu (approved):
I've added [here](https://github.com/ubiquity/ubiquity-dollar/issues/876) a reminder to revert the current PR

gitcoindev:
Now coverage job is failing, coveralls server is down or API endpoint not responding. I triggered it 3 times already with the same result..

📄 Using coverage file: packages/contracts/filtered-lcov.info
🚀 Posting coverage data to https://coveralls.io/api/v1/jobs
⚠️ Internal server error. Please contact Coveralls team.
Error: Process completed with exit code 1.

.github/workflows/codeql-analysis.yml
```diff
@@ -1,8 +1,10 @@
...
 
 on:
-  pull_request:
   push:
+    branches:
+      - development
+  # pull_request:  ## temporarily disable for pull requests until https://github.com/ubiquity/ubiquity-dollar/issues/876 is resolved.
 
 jobs:
   code-ql-analysis:
```
0x4007, line 10: Why not disable it entirely including on push? You mentioned that Code QL analysis is likely to be an issue

0x4007:
Isn't there a coverage solution we can use that doesn't require third party APIs? I'm out at an event now or else I would contribute more useful research. 

https://www.perplexity.ai/search/486e5ef2-96ef-4ac8-8845-818f8bfd0f3b

.github/workflows/codeql-analysis.yml
gitcoindev: all right let's disable completely for now

gitcoindev:
> Isn't there a coverage solution we can use that doesn't require third party APIs? I'm out at an event now or else I would contribute more useful research.
> 
> https://www.perplexity.ai/search/486e5ef2-96ef-4ac8-8845-818f8bfd0f3b

Coveralls is fixed now, this was a temporary glitch on their side. But if this happens again we could consider other options.

PR #884, opened 2024-01-24
chore: limit diamond storage check API calls

gitcoindev:
Narrow down test matrix scope to changed Dollar libraries

Resolves: #876

QA: Changed 5 files in a dummy PR, 3 files are libraries with storage, only those were picked up to be analyzed.

https://github.com/gitcoindev/ubiquity-dollar/pull/15/files#diff-3c2e3163820394bdd79dac50e9c8f83d7938e2f0b8354d5214b04b4798b114f8

GitHub diamond storage check action triggered only on 3 libraries:

https://github.com/gitcoindev/ubiquity-dollar/actions/runs/7638730341

After this pull request is accepted and merged, https://github.com/ubiquity/ubiquity-dollar/pull/878 can be reverted.

PR #885, opened 2024-01-24
Revert "chore: temporarily disable storage changes workflow for pull requests"

molecula451:
Reverts ubiquity/ubiquity-dollar#878

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,10 +1,8 @@
 name: CodeQL Analysis
 
-on: workflow_dispatch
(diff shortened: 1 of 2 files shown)


gitcoindev:
Hi @molecula451 let's revert. In the meantime I also applied the same update for core contracts on my fork as for diamond contracts, will open an another pull request soon, this should reduce GitHub action API load even more.

gitcoindev (approved):

molecula451:
@rndquu we're reverting so we have everything clean again and gitcoindev has only eyes for the new pushes at #884

PR #891, opened 2024-02-02
Update block count in a week

gitcoindev:
Resolves https://github.com/sherlock-audit/2023-12-ubiquity-judging/issues/230

Measured approximate number of blocks:

```
npx tsx scripts/task/task.ts BlocksInWeek --network=mainnet
...
Calculating number of blocks in the last week...
Recent average block time: 12 seconds
Estimated blocks in a week best case 50400
Produced 49930 blocks, 470 worst than the best case
```

QA: `Ran 33 test suites: 341 tests passed, 0 failed, 0 skipped (341 total tests)`

### DIFF
--- a/packages/contracts/scripts/task/dollar/blocks-in-week.ts
+++ b/packages/contracts/scripts/task/dollar/blocks-in-week.ts
@@ -1,0 +1,51 @@
+import { OptionDefinition } from "command-line-args";
+
+import { Networks, TaskFuncParam } from "../../shared";
+import { EtherscanProvider } from "ethers";
+
+export const optionDefinitions: OptionDefinition[] = [
+  { name: "task", defaultOption: true },
+  { name: "network", alias: "n", type: String },
+];
+
+const func = async (params: TaskFuncParam) => {
+  const { args, env } = params;
+  const { network } = args;
+
+  const chainId = Networks[network] ?? undefined;
+  if (!chainId) {
+    throw new Error(`Unsupported network: ${network} Please configure it out first`);
+  }
+
+  const provider = new EtherscanProvider(chainId, env.etherscanApiKey);
+
+  console.log(`Calculating number of blocks in the last week...`);
+  const secondsInAWeek = 604800; // 24 * 7 * 60 * 60 seconds is one week
+  const currentBlockNumber = await provider.getBlockNumber();
+  const currentBlockTimestamp = (await provider.getBlock(currentBlockNumber))?.timestamp;
+  const blockTimestampTwoBlocksAgo = (await provider.getBlock(currentBlockNumber - 2))?.timestamp;
+
+  if (currentBlockTimestamp && blockTimestampTwoBlocksAgo) {
+    const avgBlockTime = (currentBlockTimestamp - blockTimestampTwoBlocksAgo) / 2;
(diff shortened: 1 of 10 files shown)


gitcoindev:
This simple task took me much longer than I anticipated, but I thought that creating a script will enable future updates much faster. I also actually was curious to implement this. The block count is approximate, but very close to the real number. I measured 3-5 blocks difference, the exact could be achieved with iterating over the closest blocks and pulling blocks from Etherscan. Since the number fluctuates with each block and each hour I think this approximation is more than enough. The script can be applied on multiple networks.

 I tested mainnet and sepolia which revealed that on average with the same block average time which is currently 12 seconds Sepolia produced less blocks weekly, which makes sense.

rndquu (approved):

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -267,8 +267,8 @@ library LibStaking {
         // deposit new shares
         LibChef.deposit(msg.sender, _sharesAmount, _id);
         // calculate end locking period block number
-        // 1 week = 45361 blocks = 2371753*7/366
-        // n = (block + duration * 45361)
+        // 1 week = 49930 blocks
+        // n = (block number + duration * 49930)
         stake.endBlock = block.number + _weeks * ss.blockCountInAWeek;
```
molecula451, line 272: i think we should improve ALL naked math with

```solidity
unchecked { }
```

i think its best practice for upcoming updates

packages/contracts/scripts/task/dollar/blocks-in-week.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { Networks, TaskFuncParam } from "../../shared";
+import { EtherscanProvider } from "ethers";
+
+export const optionDefinitions: OptionDefinition[] = [
+  { name: "task", defaultOption: true },
+  { name: "network", alias: "n", type: String },
+];
+
+const func = async (params: TaskFuncParam) => {
```
molecula451: func is a keyword in the Go programming language, unsure why you leave it like this instead of giving a proper name according to this?

gitcoindev: hi @molecula451 I followed the naming scheme in task/dollar , see price-reset.ts . We can rename to proper names.

molecula451: yes i mean why the name has no other context? like funcBlocksinAWeek?

gitcoindev: done https://github.com/ubiquity/ubiquity-dollar/pull/891/commits/04907d7f19ddca16f6c41c9679529c43000ccb24

packages/contracts/src/dollar/libraries/LibStaking.sol
gitcoindev: @molecula451 the best would be to open a new issue, describe why it is needed and we can address this in a separate pull request.

molecula451: yes just saying

molecula451 (approved):

gitcoindev:
Thank you for reviews and merging!

PR #894, opened 2024-02-06
feat: check if collateral is enabled in collectRedemption

gitcoindev:
Also add a unit test that verifies the check.

Resolves: https://github.com/sherlock-audit/2023-12-ubiquity-judging/issues/29

### DIFF
--- a/packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
+++ b/packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
@@ -491,7 +491,11 @@ library LibUbiquityPool {
      */
     function collectRedemption(
         uint256 collateralIndex
-    ) internal returns (uint256 collateralAmount) {
+    )
+        internal
+        collateralEnabled(collateralIndex)
+        returns (uint256 collateralAmount)
+    {
         UbiquityPoolStorage storage poolStorage = ubiquityPoolStorage();
 
         require(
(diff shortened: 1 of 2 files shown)


rndquu (approved):

rndquu:
This is wrong to push all fixes (low and medium/high) to the same `fix-review` branch according to the sherlock rules but we've already merged a couple of them. Chances are we will have to cherry pick PRs again when the sherlock's judging stage is over.

molecula451 (approved):
the branch `fix-review` is full of commits already

packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
```diff
@@ -491,7 +491,11 @@ library LibUbiquityPool {
      */
     function collectRedemption(
         uint256 collateralIndex
-    ) internal returns (uint256 collateralAmount) {
+    )
```
0xadrii, line 494: [MEDIUM] - Adding collateral enabled checks for redemption collections might make users uncapable of redeming their funds.

This PR is intended to fix [the bug described in this Sherlock issue](https://github.com/sherlock-audit/2023-12-ubiquity-judging/issues/29). However, the Sherlock report is actually wrong and the vulnerability described must not be considered as an issue in the first place, so the collateralEnabled check should not be added on collecting redemptions.

Enabling/disabling collaterals allows the protocol to add or remove supported collaterals, not to be used under uncertain situations, such as a hack. In such type of situations, the `isRedeemPaused` check can be enabled to prevent users from redeeming if something wrong is currently  happening.

The problem with adding the `collateralEnabled` check here is that if Ubiquity decides to disable a collateral in the future, some redemptions might be already queued, so users won't be able to withdraw their funds.


Recommendation: Remove the `collateralEnabled` check for collecting redemptions.

PR #904, opened 2024-02-25
feat: protect protocol upgradeability loss

### LINKED ISSUE #21: chaduke - The owner might remove selector DiamondCutFacet.diamondCut() and then the protocol loses upgradablity.
chaduke

medium

# The owner might remove selector DiamondCutFacet.diamondCut() and then the protocol loses upgradablity.

## Summary
The owner might remove selector DiamondCutFacet.diamondCut() and then the protocol loses upgradablity.

## Vulnerability Detail
The Ubiquity protocol uses the Diamond proxy to support upgradability of the protocol. There are two cases to add a facet:
1) When Diamond is first deployed, facets can be added via the constructor, in which the DiamondCutFacet and its selectors are added as well; 2) future facets cut are accomplished via DiamondCutFacet.diamondCut().

Therefore, it is important that DiamondCutFacet.diamondCut() will not be removed via itself. However, the following POC shows that the owner (either malicious, or stolen or by accident) might remove DiamondCutFacet.diamondCut().

```javascript
   function testDiamondCutFacet1() public {
        FacetCut[] memory facetCut = new FacetCut[](1);
        facetCut[0] = FacetCut({
            facetAddress: address(0),
            action: FacetCutAction.Remove,
            functionSelectors: selectorsOfDiamondCutFacet
        });

        vm.prank(owner);
        diamondCutFacet.diamondCut(facetCut, address(0x0), "");
    }
```

The following code shows that after removing, it is impossible to add it back due to the failing of calling diamondCutFacet.diamondCut(). The error message is that "the function does not exists"

```javascript
   function testDiamondCutFacet2() public {
        FacetCut[] memory facetCut1 = new FacetCut[](1);
        facetCut1[0] = FacetCut({
            facetAddress: address(0),
            action: FacetCutAction.Remove,
            functionSelectors: selectorsOfDiamondCutFacet
        });

        vm.prank(owner);
        diamondCutFacet.diamondCut(facetCut1, address(0x0), "");

        FacetCut[] memory facetCut2 = new FacetCut[](1);
        facetCut2[0] = FacetCut({
            facetAddress: address(diamondCutFacet),
            action: FacetCutAction.Add,
            functionSelectors: selectorsOfDiamondCutFacet
        });
        vm.prank(owner);
        vm.expectRevert();
        diamondCutFacet.diamondCut(facetCut2, address(0x0), "");
    }

```

Note the following line prevents the removel of a function from diamond.sol but not from DiamondCutFacet, which has a different address from Diamond (the proxy): 

[https://github.com/sherlock-audit/2023-12-ubiquity/blob/d9c39e8dfd5601e7e8db2e4b3390e7d8dff42a8e/ubiquity-dollar/packages/contracts/src/dollar/libraries/LibDiamond.sol#L326](https://github.com/sherlock-audit/2023-12-ubiquity/blob/d9c39e8dfd5601e7e8db2e4b3390e7d8dff42a8e/ubiquity-dollar/packages/contracts/src/dollar/libraries/LibDiamond.sol#L326)

## Impact
The owner might remove selector DiamondCutFacet.diamondCut() and then the protocol loses upgradablity.

## Code Snippet

[https://github.com/sherlock-audit/2023-12-ubiquity/blob/d9c39e8dfd5601e7e8db2e4b3390e7d8dff42a8e/ubiquity-dollar/packages/contracts/src/dollar/facets/DiamondCutFacet.sol#L14-L21](https://github.com/sherlock-audit/2023-12-ubiquity/blob/d9c39e8dfd5601e7e8db2e4b3390e7d8dff42a8e/ubiquity-dollar/packages/contracts/src/dollar/facets/DiamondCutFacet.sol#L14-L21)

## Tool used
Foundry

Manual Review

## Recommendation
Add code into  DiamondCutFacet.diamondCut() to check that  the argument ``_diamondCut`` does not contain the cut that will remove DiamondCutFacet.diamondCut().

gitcoindev:
diamondCut function cannot be removed.

Resolves: https://github.com/sherlock-audit/2023-12-ubiquity-judging/issues/21

### DIFF
--- a/packages/contracts/src/dollar/libraries/LibDiamond.sol
+++ b/packages/contracts/src/dollar/libraries/LibDiamond.sol
@@ -321,7 +321,8 @@ library LibDiamond {
             _facetAddress != address(0),
             "LibDiamondCut: Can't remove function that doesn't exist"
         );
-        // precomputed bytes4(keccak256()) diamondCut function selector 0x1f931c1c to save gas
+        // precomputed diamondCut function selector to save gas
+        // bytes4(keccak256(abi.encodeWithSignature("diamondCut((address,uint8,bytes4[])[],address,bytes)"))) == 0x1f931c1c
         require(
             _selector != bytes4(0x1f931c1c),
             "LibDiamondCut: Can't remove diamondCut function"
--- a/packages/contracts/src/dollar/libraries/LibDiamond.sol
+++ b/packages/contracts/src/dollar/libraries/LibDiamond.sol
@@ -321,6 +321,11 @@ library LibDiamond {
             _facetAddress != address(0),
             "LibDiamondCut: Can't remove function that doesn't exist"
         );
+        // precomputed bytes4(keccak256()) diamondCut function selector 0x1f931c1c to save gas
+        require(
+            _selector != bytes4(0x1f931c1c),
+            "LibDiamondCut: Can't remove diamondCut function"
+        );
         // an immutable function is a function defined directly in a diamond
         require(
             _facetAddress != address(this),
--- a/packages/contracts/test/diamond/DiamondTest.t.sol
+++ b/packages/contracts/test/diamond/DiamondTest.t.sol
@@ -287,6 +287,23 @@ contract TestDiamond is DiamondTestSetup {
         IMockFacet(address(diamondCutFacet)).functionB();
     }
 
+    function testCutFacetShouldNotRemoveDiamondCutFunction() public {
+        FacetCut[] memory facetCut = new FacetCut[](1);
+        bytes4[] memory selectors = new bytes4[](1);
+        selectors[0] = diamondCutFacet.diamondCut.selector;
+
+        facetCut[0] = FacetCut({
+            facetAddress: address(0),
+            action: FacetCutAction.Remove,
+            functionSelectors: selectors
+        });
+
+        // try to remove diamondCut function
+        vm.prank(owner);
+        vm.expectRevert("LibDiamondCut: Can't remove diamondCut function");
+        diamondCutFacet.diamondCut(facetCut, address(0x0), "");
+    }
+
     function testSelectors_ShouldBeAssociatedWithCorrectFacet() public {
         for (uint256 i; i < facetAddressList.length; i++) {
             if (compareStrings(facetNames[i], "DiamondCutFacet")) {


packages/contracts/src/dollar/libraries/LibDiamond.sol
```diff
@@ -321,6 +321,11 @@ library LibDiamond {
             _facetAddress != address(0),
             "LibDiamondCut: Can't remove function that doesn't exist"
         );
+        // precomputed bytes4(keccak256()) diamondCut function selector 0x1f931c1c to save gas
+        require(
+            _selector != bytes4(0x1f931c1c),
```
rndquu, line 327: @gitcoindev  Can we make it explicit without hard to read magic bytes?

0x4007: Well the comment is pretty clear for what it does and why its there. I think its probably fine unless there is a better way to do this.

gitcoindev: I am aware this looks like black magic, that's why I added the comment to clarify where did this value came from. I also run the test case locally with `-vvv` debug to confirm the value in the logs. The best explanation I found was at https://solidity-by-example.org/function-selector/

rndquu: > I am aware this looks like black magic, that's why I added the comment to clarify where did this value came from. I also run the test case locally with `-vvv` debug to confirm the value in the logs. The best explanation I found was at https://solidity-by-example.org/function-selector/

1. `require(_selector != DiamondCutFacet.diamondCut.selector)` is way more explicit
2. I have to spend time to check that `bytes4(keccak256(abi.encodeWithSignature("diamondCut((address,uint8,bytes4[])[],address,bytes)"))) == 0x1f931c1c` (which is `true` by the way)
3. Next time some contributor will see these magic bytes and will think that "ohh these guys like gas optimizations, why don't I introduce a completely unaudited custom bit math lib"

Approving this because you're busy with staking which is way more important that arguing over 4 bytes

rndquu (approved):

packages/contracts/src/dollar/libraries/LibDiamond.sol
gitcoindev: Thank you @rndquu for your effort and valuable feedback. I am going to add 2) to the comment, it will be clear for everyone else.

rndquu: Thank you

molecula451 (approved):

PR #920, opened 2024-03-25
refactor(migration): deploy collateral in testnet

rndquu:
This PR refactors migration script:
- for mainnet we use `LUSD` collateral token (read its address from env variable)
- for testnet/anvil we deploy collateral token from scratch (for ease of debugging)


molecula451 (approved):
no issues so far, just a quick update

gitcoindev (commented):
Reviewed after merging, was driving a long trip yesterday. Looks good!

PR #922, opened 2024-04-04
Feat/fractional pool

rndquu:
This PR introduces fractional mint/redeem of the Dollar token based on the collateral ratio which represents how many tokens user has to provide in collateral and governance tokens.

How it worked before:
- mint: user provides 1 LUSD and gets 1 Dollar token
- redeem: user provides 1 Dollar token and gets 1 LUSD

**How it works in this PR**

Example 1 (collateral ratio is 100%):
- mint: user provides 1 LUSD and gets 1 Dollar token
- redeem: user provides 1 Dollar token and gets 1 LUSD

You may see that the behaviour is the same as in the previous code version when collateral ratio is 100%.

Example 2 (collateral ratio is 90%):
- mint: user provides 0.9 LUSD + 0.1 USD in UBQ tokens and gets 1 Dollar token
- redeem: user provides 1 Dollar token and gets 0.9 LUSD + 0.1 USD in UBQ tokens

The main reason why we need the fractional feature from the start is to incentivise initial investors (who staked collateral to get UBQ tokens) to put liquidity in the protocol. The plan is to set collateral ratio to 95% from the start so that initial investors could mint Dollar tokens with a 5% discount which is both profitable for investors and useful for protocol (i.e. Dollar token) liquidity.

The code works the same way as in one of the [popular stablecoins](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Pools/FraxPoolV3.sol#L373-L387).

Notice:
1. We let the admin to set the collateral ratio value while the popular stablecoin protocol uses [dynamic collateral ratio](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Frax.sol#L197-L221). Dynamic collateral ratio works this way: if stablecoin USD price is >$1.00 for some time then global collateral ratio is dicreased, on the contrary if stablecoin USD price is <$1.00 for some time then global collateral ratio is increased. We omit this mechanic for simplicity.
2. There is also the mechanic of [decollateralization](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Pools/FraxPoolV3.sol#L499) and [recollateralization](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Pools/FraxPoolV3.sol#L534) which works this way:
a) If collateral ratio is increased from 90% to 100% (dynamically or admin set it) then the pool contains 10% of missing collateral so users can deposit collateral to the pool in exchange for Governance tokens (with some premium) until the actual collateral value hits 100%. It is called recollateralization.
b) On the contrary, in case of decollateralization, if collateral ratio is decreased from 100% to 90% then the pool contains 10% of excess collateral. Users can burn their Governance tokens in exchange for collateral until the actual collateral value is 90%.
Initially we plan to set collateral ratio to 95% and leave it at that level. Besides users can [override](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Pools/FraxPoolV3.sol#L360) fractional mechanic on mint. But should we want to update collateral ratio from 95% to some other value after the deployment we'll have to add decollateralization and recollateralization methods to the protocol so that users could rebalance the pool.

P.S. Sadly this PR modifies already audited contracts so we'll have to find somebody from the Sherlock's audit to check the updates
P.P.S Diamond storage check workflow is [failing](https://github.com/ubiquity/ubiquity-dollar/actions/runs/8557216549/job/23448939640?pr=922) which is expected and is ok since the contracts are not deployed yet.
P.P.P.S. I will update mainnet and development deploy script in another PR

### DIFF
--- a/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
+++ b/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
@@ -33,6 +33,11 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
     }
 
     /// @inheritdoc IUbiquityPool
+    function collateralRatio() external view returns (uint256) {
+        return LibUbiquityPool.collateralRatio();
+    }
+
+    /// @inheritdoc IUbiquityPool
     function collateralUsdBalance()
         external
         view
@@ -178,6 +183,11 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
             chainLinkPriceFeedAddress,
             stalenessThreshold
         );
+    }
+
+    /// @inheritdoc IUbiquityPool
+    function setCollateralRatio(uint256 newCollateralRatio) external onlyAdmin {
+        LibUbiquityPool.setCollateralRatio(newCollateralRatio);
     }
 
     /// @inheritdoc IUbiquityPool
--- a/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
+++ b/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
@@ -84,6 +84,15 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
     }
 
     /// @inheritdoc IUbiquityPool
+    function getGovernancePriceUsd()
+        external
+        view
+        returns (uint256 governancePriceUsd)
+    {
+        return LibUbiquityPool.getGovernancePriceUsd();
+    }
+
+    /// @inheritdoc IUbiquityPool
     function getRedeemCollateralBalance(
         address userAddress,
         uint256 collateralIndex
--- a/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
+++ b/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
@@ -86,6 +86,11 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
             );
     }
 
+    /// @inheritdoc IUbiquityPool
+    function governanceEthPoolAddress() external view returns (address) {
+        return LibUbiquityPool.governanceEthPoolAddress();
+    }
+
     //====================
     // Public functions
     //====================
@@ -197,6 +202,15 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
         uint256 newRedeemFee
     ) external onlyAdmin {
         LibUbiquityPool.setFees(collateralIndex, newMintFee, newRedeemFee);
+    }
+
+    /// @inheritdoc IUbiquityPool
+    function setGovernanceEthPoolAddress(
+        address newGovernanceEthPoolAddress
+    ) external onlyAdmin {
+        LibUbiquityPool.setGovernanceEthPoolAddress(
+            newGovernanceEthPoolAddress
+        );
     }
 
     /// @inheritdoc IUbiquityPool
--- a/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
+++ b/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
@@ -171,7 +171,11 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
     /// @inheritdoc IUbiquityPool
     function collectRedemption(
         uint256 collateralIndex
-    ) external nonReentrant returns (uint256 collateralAmount) {
+    )
+        external
+        nonReentrant
+        returns (uint256 governanceAmount, uint256 collateralAmount)
+    {
         return LibUbiquityPool.collectRedemption(collateralIndex);
     }
 
--- a/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
+++ b/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
@@ -105,6 +105,13 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
     }
 
     /// @inheritdoc IUbiquityPool
+    function getRedeemGovernanceBalance(
+        address userAddress
+    ) external view returns (uint256) {
+        return LibUbiquityPool.getRedeemGovernanceBalance(userAddress);
+    }
+
+    /// @inheritdoc IUbiquityPool
     function governanceEthPoolAddress() external view returns (address) {
         return LibUbiquityPool.governanceEthPoolAddress();
     }
@@ -145,12 +152,18 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
     function redeemDollar(
         uint256 collateralIndex,
         uint256 dollarAmount,
+        uint256 governanceOutMin,
         uint256 collateralOutMin
-    ) external nonReentrant returns (uint256 collateralOut) {
+    )
+        external
+        nonReentrant
+        returns (uint256 collateralOut, uint256 governanceOut)
+    {
         return
             LibUbiquityPool.redeemDollar(
                 collateralIndex,
                 dollarAmount,
+                governanceOutMin,
                 collateralOutMin
             );
     }
--- a/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
+++ b/packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
@@ -118,18 +118,26 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
         uint256 collateralIndex,
         uint256 dollarAmount,
         uint256 dollarOutMin,
-        uint256 maxCollateralIn
+        uint256 maxCollateralIn,
+        uint256 maxGovernanceIn,
+        bool isOneToOne
     )
         external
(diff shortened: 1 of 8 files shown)


packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
```diff
@@ -87,6 +89,15 @@ library LibUbiquityPool {
         bool[] isMintPaused;
         // whether redeeming is paused for a particular collateral index
         bool[] isRedeemPaused;
+        //====================================
+        // Governance token pricing related
+        //====================================
+        // chainlink price feed for ETH/USD pair
+        address ethUsdPriceFeedAddress;
+        // threshold in seconds when chainlink's ETH/USD price feed answer should be considered stale
+        uint256 ethUsdPriceFeedStalenessThreshold;
```
gitcoindev, line 103: I do not see `ethUsdPriceFeedStalenessThreshold` being used, could you please also add a test for this that verifies behavior of stale price feed? Perhaps I am to tired today but I just see setters/ getters tests and an emitted event only.

rndquu: This PR is a draft

packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
```diff
@@ -69,6 +83,15 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
         return LibUbiquityPool.getDollarPriceUsd();
     }
 
+    /// @inheritdoc IUbiquityPool
+    function getGovernancePriceUsd()
```
molecula451, line 87: please lets do USD, all upper case

packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
```diff
@@ -41,6 +46,15 @@ contract UbiquityPoolFacet is IUbiquityPool, Modifiers {
         return LibUbiquityPool.collateralUsdBalance();
     }
 
+    /// @inheritdoc IUbiquityPool
+    function ethUsdPriceFeedInformation()
```
molecula451, line 50: `ethUSDPriceFeedInformation()` looks equally better

packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
0x4007: According to our code styleguide it's supposed to be `Usd`

I guess it's not explicitly stated but it's implicit due to the `Nft` [rule](https://github.com/ubiquity/ubiquity-dollar/blob/development/.github/CONTRIBUTING.md)

molecula451: i think it's fine too tho, altho usually tickers symbols are always uppercase, what you think @gitcoindev

molecula451 (approved):
We are tight on time, this PR looks good to me

packages/contracts/src/dollar/facets/UbiquityPoolFacet.sol
gitcoindev: Let's use the style guide then.

gitcoindev (approved):

packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
```diff
@@ -357,68 +477,106 @@ library LibUbiquityPool {
...
+                collateralIndex,
+                dollarForCollateral
+            );
+            governanceNeeded = dollarForGovernance
+                .mul(UBIQUITY_POOL_PRICE_PRECISION)
+                .div(getGovernancePriceUsd());
+        }
 
         // subtract the minting fee
         totalDollarMint = dollarAmount
```
0xadrii, line 549: [HIGH] - Substracting the minting fee after computing the collateral needed will break dollar stability.

The current approach to mint is (considering the simplest scenario where CR > 100% or `isOneToOne` is set to `true`):
1. User calls `mintDollar` and specifies the amount of dollars he wants to mint via the `dollarAmount` parameter
2. The collateral needed is computed from the `dollarAmount` requested by the user
3. **After** computing the collateral needed to mint `dollarAmount`, the `mintingFee` is substracted from `dollarAmount` so that only `totalDollarMint` will be minted

This approach is wrong and breaks the peg mechanism, given that the amount of dollar minted will always be smaller than the expected by the peg algorithm. Consider the following numbers (assuming 1 dollar = 1 LUSD):
1. User wants to mint dollar and specifies a  `dollarAmount` of 1000.
2. `getDollarInCollateral` returns a `collateralNeeded` of 1000
3. After that, the fee is substracted from `dollarAmount` (let's say a 10% applies) and the final `totalDollarMint` amount to mint is 900 dollar.

The result is we have dollar always overbacked by collateral, which will draw dollar price down because every time dollar is minted the supply becomes more scarce than the amount backing it.

This situation is similar to adding fees to an ERC4626 vault and calling the `previewMint`. As [OpenZeppelin's research](https://docs.openzeppelin.com/contracts/4.x/erc4626#fees) shows, when the shares are requested, the fees are added on top of the computed assets (not substracted from them), so that this imbalance does not occur.


On the other hand, the `redeemDollar` flow is correct. When redeeming, users will specify the `dollarAmount` they want to redeem, and the first action taken will be to substract the fee (`dollarAfterFee`). Then, all computations will be performed considering the already-substracted fee, so no supply imbalance will take place.

rndquu, line 549: Original Frax contract [works](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Pools/FraxPoolV3.sol#L390) the same way.

> This approach is wrong and breaks the peg mechanism, given that the amount of dollar minted will always be smaller than the expected by the peg algorithm

The peg mechanism is based on arbitrage opportunity (users must always be able to mint/redeem 1 Dollar token for exactly 1 USD in collateral), not on the amount of collateral.

> The result is we have dollar always overbacked by collateral, which will draw dollar price down

I don't understand why overcollateralization is bad. On the contrary, the greater the collateral/Dollar ratio the better the pool's resistance to collateral depeg.

Overall I think that overcollateralization doesn't harm the protocol and USD peg in particular.

@0xadrii Are there any similar issues on sherlock/code4arena ?

@gitcoindev @molecula451 Help

gitcoindev: I also always considered overcollateralization as good. Most protocols are overcollateralized by design. For LUSD (Liquity) which is also considered as current main collateral for Dollar, MCR (minimal collateral ratio) is 110%.
User deposits ETH, which is 110% value of LUSD, gets LUSD backed loan. 

https://medium.com/@lefl0wdutemp/a-deepdive-into-liquity-v2-7463beff53d4 is a good explanation.

I will have to give some more thought to @0xadrii 's findings, and read the OpenZeppelin's research as well.

0xadrii: Yes, just checked a bit deeper and it does not have the impact I initially thought. I found a similar issue in another project but the internal mechanics were a bit different so it affected the protocol in a different way, so in this case it is a false positive.

packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
```diff
@@ -778,6 +992,45 @@ library LibUbiquityPool {
...
+     * - Mint: user provides 1 collateral token to get 1 Dollar
+     * - Redeem: user gets 1 collateral token for 1 Dollar
+     *
+     * @dev Example (900_000 = 90%):
+     * - Mint: user provides 0.9 collateral token and 0.1 Governance token to get 1 Dollar
+     * - Redeem: user gets 0.9 collateral token and 0.1 Governance token for 1 Dollar
+     *
+     * @param newCollateralRatio New collateral ratio
+     */
+    function setCollateralRatio(uint256 newCollateralRatio) internal {
```
0xadrii, line 1009: [LOW] - Collateral ratio should be capped

As seen in [Frax's code](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Frax.sol#L212), collateral ratio will always be capped at 100%. For Ubiquity, if a higher value is wrongly set and the dynamic CR is implemented, then it could cause some issues. It is recommended to add a maximum value for `newCollateralRatio` of 1_000_000.

It would also be good to cap the fees in the `setFees` function to the max acceptable values, as well as the redemption delay blocks in `setRedemptionDelayBlocks`.

PR #923, opened 2024-04-08
chore: some updates, cast to uint256

molecula451:
-  We are casting anything from `uint` to `uint256`, this is important everywhere, also `uint` it's an alias for `uint256`, `uint`  it's gonna be deprecated in newer solidity compilers
- There is not reason to keep using https://github.com/ubiquity/ubiquity-dollar/blob/91816753d250dfe34f6956eb7fc6d66a5ce0e96c/packages/contracts/deployments.json, confusing contributors and the enhance of the repo, why,  because we rely on run-latest.json usually by forge build

I was  checking to  see if SMT break  somewhere, looks working still (locally)


packages/contracts/scripts/smt-checker/README.md
```diff
@@ -12,7 +12,7 @@ Ensure that your system has Z3 installed with a version number greater than or e
 Before proceeding, ensure that you have `g++` and `python3` installed on your system. To download and install Z3, enter the following commands in your terminal:
 
 ### Run script:
-```
+```sh
 sh ./update-dependencies.sh
```
gitcoindev: The scripts were renamed:
```
./check-deps.sh
./update-deps-ubuntu.sh
```
It should also reflect the README content. Also sh can be removed, as the shell scripts have executable bit set and can be executed directly from shell:
```
$ ./check-deps.sh 
All required programs are installed on your system.
$ ./update-deps-ubuntu.sh 
```

gitcoindev (changes requested):

packages/contracts/scripts/smt-checker/README.md
molecula451: indeed

gitcoindev (approved):

rndquu (approved):

PR #932, opened 2024-04-20
feat: cap collateral ratio to 100%

### LINKED ISSUE #929: [LOW] - Collateral ratio should be capped
[LOW] - Collateral ratio should be capped

As seen in [Frax's code](https://github.com/FraxFinance/frax-solidity/blob/967002fc7f2c5ee9e175cabc763a776a8ed03e01/src/hardhat/contracts/Frax/Frax.sol#L212), collateral ratio will always be capped at 100%. For Ubiquity, if a higher value is wrongly set and the dynamic CR is implemented, then it could cause some issues. It is recommended to add a maximum value for `newCollateralRatio` of 1_000_000.

_Originally posted by @0xadrii in https://github.com/ubiquity/ubiquity-dollar/pull/922#discussion_r1572505552_

gitcoindev:
Resolves: #929

### DIFF
--- a/packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
+++ b/packages/contracts/src/dollar/libraries/LibUbiquityPool.sol
@@ -994,7 +994,8 @@ library LibUbiquityPool {
 
     /**
      * @notice Sets collateral ratio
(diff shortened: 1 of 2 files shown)


rndquu (approved):

molecula451 (approved):

0xadrii:
Fix looks good

PR #941, opened 2024-06-20
chore: add initial jest for testing UI

molecula451:
Resolves #


0x4007:
I thought we use cypress for ui tests

molecula451:
> I thought we use cypress for ui tests

Indeed we are way behind on UI stuff, the infra is set, say contracts, we could trade between this both cypress/jest to increase branch robustness

PR #947, opened 2024-06-26
chore: RPC update

### LINKED ISSUE #946: Update RPC Endpoint on Foundry Config & Readme
Currently the RPC Endpoint in our readme as far as I concern it was a "forwarded" route to an external/public endpoint that do not belong to us (Ubiquity Org/Team)

The endpoint it's also not available

```curl
curl -X POST -H "Content-Type: application/json" --data '{
    "jsonrpc":"2.0",
    "method":"eth_getStorageAt",
    "params":[
        "0x1",
        "0xfff",
        "latest"
    ],
    "id":1
}' https://eth.ubq.fi/v1/mainnet
```

the curl call will return incorrect data:

**Origin DNS Error**

image

Replace with any other public available endpoint

molecula451:
Resolves #946 

This is only a public mainnet RPC gateway & it's own by ankr. If we do have better suggestion then I think we should add it up until if we ever run our own dedicated node

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ yarn workspace @ubiquity/dapp start # Run the web application at http://localhos
 ### Network Settings
 | Network | Chain ID | RPC Endpoint                  | Comment |
 |---------|----------|-------------------------------|---------|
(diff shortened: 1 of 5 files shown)


rndquu (approved):

gitcoindev (approved):
Reviewed and approving.

PR #956, opened 2024-09-07
feat: add an AMO minter and an Aave AMO

### LINKED ISSUE #611: Algorithmic Market Operations for Collateral

zugdev:
Resolves #611


0x4007 (changes requested):
Refer to our other smart contracts for code styles. 

Never reference token symbols as identifiers. Besides we already renamed uad to uusd, causing developer confusion down the road. 

uAD -> Dollar
UBQ -> Governance

packages/contracts/LICENSE_GPL
0x4007, line 1: Is this considered the best practice?

zugdev: It is mandatory to clearly reference source code under GNU license. This is a common way to do that.

zugdev:
- @0x4007 @rndquu can I get some feedback? I will start writing tests for the AMO minter and for the Aave AMO, therefore any core feedback would be invaluable until then. Regarding code style I will adhere to it as a last step, don't worry about it.

packages/contracts/src/dollar/libraries/TransferHelper.sol
0x4007, line 1: Don't we already have our own safe transfer methods built-in to our own tokens?

0x4007 (commented):
I'm not a solidity developer, but nothing really stands out except for my one comment

rndquu:
> * @0x4007 @rndquu can I get some feedback? I will start writing tests for the AMO minter and for the Aave AMO, therefore any core feedback would be invaluable until then. Regarding code style I will adhere to it as a last step, don't worry about it.

Hey, sorry for a late response, I'll check it today

packages/contracts/src/dollar/libraries/TransferHelper.sol
zugdev: Thanks, will check that

rndquu:
@zugdev 

Notes:
1. [AaveAMO](https://github.com/ubiquity/ubiquity-dollar/blob/96e72d710bfea2dbe88938410cede5743b2b7964/packages/contracts/src/dollar/amo/AaveAMO.sol) doesn't really use [UbiquityAMOMinter](https://github.com/ubiquity/ubiquity-dollar/blob/96e72d710bfea2dbe88938410cede5743b2b7964/packages/contracts/src/dollar/core/UbiquityAMOMinter.sol) (although defines it in the constructor) so it's not clear how those 2 contract will interact with each other and `UbiquityPool`.
2. [UbiquityAMOMinter](https://github.com/ubiquity/ubiquity-dollar/blob/96e72d710bfea2dbe88938410cede5743b2b7964/packages/contracts/src/dollar/core/UbiquityAMOMinter.sol) is basically a fork of [FraxAMOMinter](https://github.com/FraxFinance/frax-solidity/blob/85039d4dff2fb24d8a1ba6efc1ebf7e464df9dcf/src/hardhat/contracts/Frax/FraxAMOMinter.sol) (which is expected). We need only methods related to collateral moving to and from AMO strategies.
3. All [aave](https://github.com/ubiquity/ubiquity-dollar/tree/96e72d710bfea2dbe88938410cede5743b2b7964/packages/contracts/src/dollar/interfaces/aave) related structs/interfaces/contracts should be taken from `aave` github repository (i.e. `aave` should be installed in [github submodules](https://github.com/ubiquity/ubiquity-dollar/blob/0a230f977ad91bbfb01dd28bc1d2ea77faed5237/.gitmodules)) instead of copy/pasting them.
4. There is no need in copy/pasting [TransferHelper](https://github.com/ubiquity/ubiquity-dollar/blob/96e72d710bfea2dbe88938410cede5743b2b7964/packages/contracts/src/dollar/libraries/TransferHelper.sol) lib, just use [SafeERC20](https://github.com/ubiquity/ubiquity-dollar/blob/0a230f977ad91bbfb01dd28bc1d2ea77faed5237/packages/contracts/src/dollar/libraries/LibUbiquityPool.sol#L23). 

Question:
- What contract did you use as a base one from the Frax [github repository](https://github.com/FraxFinance/frax-solidity) for the [AaveAMO](https://github.com/ubiquity/ubiquity-dollar/blob/96e72d710bfea2dbe88938410cede5743b2b7964/packages/contracts/src/dollar/amo/AaveAMO.sol) contract?

zugdev:
1. and 2. - was waiting for this specific clarification to resolve.

3. and 4. - understood.

> What contract did you use as a base one from the Frax repo for the AaveAMO

Used [AaveAMO_V2](https://github.com/FraxFinance/frax-solidity/blob/master/src/hardhat/contracts/Misc_AMOs/Lending_AMOs/AaveAMO_V2.sol) 

Yes, there is still some stuff to be tuned, will wrap it this week.

zugdev:
Yo it should be well directed now. I've:

1. Refactored Aave AMO from V2 to V3, most liquidity is there. 
2. Implemented rewards collection mechanism.
3. Using contracts from git submodule instead of copypasted interfaces.
4. Refactored code to use collateral only.
5. Implemented emergency mechanisms: arbitrary calls and ERC20 recovery for owner - your choice to keep.

The AMO has the minter address to return collateral. I have implemented *borrow* and *repay* as well, this could be kept or removed - its a trust decision.

I still have to write tests and addapt code style, will be pushing this when I get some free time.

A little more context on 5: to handle rewards, borrows and repays, the system would need some way to swap, could design a Uniswap AMO. There could already be something in place, I haven't checked if it exists as a facet.

0x4007:
Cowswap is preferred if that's an option!

zugdev:
> Cowswap is preferred if that's an option!

Can do that! Let me wrap this one first though.

packages/contracts/test/amo/AaveAMO.t.sol
```diff
@@ -0,0 +1,331 @@
...
+    address constant newAMOMinterAddress = address(5); // mock new AMO minter address
+    address constant nonAMO = address(9999); // Address representing a non-AMO entity
+    uint256 constant interestRateMode = 2; // Variable interest rate mode in Aave
+
+    // Mocking the Aave Pool
+    IPool private constant aave_pool =
+        IPool(0x6Ae43d3271ff6888e7Fc43Fd7321a503ff738951); // Aave V3 Sepolia Pool
+
+    function setUp() public override {
+        vm.createSelectFork(vm.rpcUrl("sepolia"));
```
rndquu: The whole codebase uses plain unit tests. We used to have tests run in a forked forge instance but:
1. They tend to fail because of RPC errors
2. They run longer compared to unit tests
3. It's more clear to have everything deployed from scratch in your tests compared to using already deployed contracts which is kind of a "black box"

Can we deploy aave contracts ourselves and use plain unit tests?

zugdev: I understand where you come from. This is a speed/quality balance, will try to match this pattern - however deploying Aave is not that trivial.

0x4007: Couldn't our rpc manager help with the fail in requests problem?

rndquu: > I understand where you come from. This is a speed/quality balance, will try to match this pattern - however deploying Aave is not that trivial.

If deploying Aave pool from scratch takes too much time then you may test against a mocked instance. We'll refactor later.

> Couldn't our rpc manager help with the fail in requests problem?

Testing against a forked instance means testing against a predefined state. It's way better when you fully understand the protocol where you send assets to (i.e. when you deploy aave yourself) compared to forked tests where you use already deployed addresses of aave contracts, addresses of users who have funds, addresses of external protocols, etc...

rndquu: > I understand where you come from. This is a speed/quality balance, will try to match this pattern - however deploying Aave is not that trivial.

@zugdev  I think you may keep forked tests (because deploying aave from scratch was not stated in the original issue) but we'll have to refactor them later. Can't we fork from mainnet instead of sepolia?

zugdev: I'd be down to refactor it later, but as you've said it will take time. Regarding protocol understanding, I fully understand how Aave operates, as I've forked it many times now - but I do understand the importance of consistency in a large codebase. The main advantage of forking sepolia instead of mainnet is that we minimize the number of "magic addresses" we prank or use, paired with the permission to change protocol configs. Aave in sepolia is prepared to be tested on, they support mintable tokens for example. Another big advantage is interest rate and fees are set to very low, which allow us to assert that the deposit amount is equal to aToken balance post-deposit. 

It is up to you what chain we fork from, but I'd recommend Sepolia. You can further read why at https://docs.aave.com/developers/guides/testing-guide

rndquu: > It is up to you what chain we fork from, but I'd recommend Sepolia. You can further read why at https://docs.aave.com/developers/guides/testing-guide

It doesn't answer the question why Sepolia is recommended. 

> Aave in sepolia is prepared to be tested on

That's why tests should use a real production mainnet environment (although forked).

Ok, in order to save time let's stick to sepolia keeping in mind that all aave instances will be deployed from scratch in unit tests later.

zugdev:
@rndquu I've marked this PR as ready for review. Please let me know if it is up to your standards. I agree an issue to write a non-forked test suite is appropriate, I would be down to go through with it.

packages/contracts/src/dollar/interfaces/IAMO.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.6.11;
```
rndquu: It's better to make it consistent with the code base:
```
pragma solidity 0.8.19;
```

rndquu (changes requested):
Ok, here comes a list of code style refactors:
1. The whole codebase uses camel cases naming so all snake case variables should be refactored to camel case
2. There's also a convention across the whole code base to use camel case for abbreviations (like `Nft` not `NFT`). So all `AMO` related namings should be refactored to `Amo`. Examples:
  - `receiveCollatFromAMO => receiveCollatFromAmo`
  - `validAMO => validAmo`
3. All `UAD` namings should be refactored to `Dollar`
4. No shortages. Be as explicit as possible. Example:
  - `receiveCollatFromAMO => receiveCollateralFromAMO`

Regarding the code:
- [TransferHelper](https://github.com/ubiquity/ubiquity-dollar/blob/30c95127f557ea02fe5c1b87f85023e3ce218159/packages/contracts/src/dollar/libraries/TransferHelper.sol) lib should removed and [SafeERC20](https://github.com/ubiquity/ubiquity-dollar/blob/0a230f977ad91bbfb01dd28bc1d2ea77faed5237/packages/contracts/src/dollar/libraries/LibUbiquityPool.sol#L23) should be used. Otherwise we'll have to perform a new audit contest solely for the copy/pasted lib.

packages/contracts/src/dollar/interfaces/IAMO.sol
zugdev: Done! I've also added interfaces to the contracts developed. This includes IAmo which sets a standard of what functions an AMO must implement. This includes returning collateral to minter and setting the AMO minter address.

.gitmodules
```diff
@@ -37,4 +37,10 @@
...
 	url = https://github.com/smartcontractkit/chainlink-brownie-contracts
-	branch= main
\ No newline at end of file
+	branch= main
+[submodule "packages/contracts/lib/aave-v3-core"]
+	path = packages/contracts/lib/aave-v3-core
+	url = https://github.com/aave/aave-v3-core
+[submodule "packages/contracts/lib/aave-v3-periphery"]
+	path = packages/contracts/lib/aave-v3-periphery
+	url = https://github.com/aave/aave-v3-periphery
```
rndquu, line 47: Could you specify the branches used? ([example](https://github.com/ubiquity/ubiquity-dollar/blob/02bbe8f425bca7d720e4d0599a86eb28b25a77f4/.gitmodules#L28))

packages/contracts/src/dollar/amo/AaveAmo.sol
```diff
@@ -0,0 +1,272 @@
...
+        emit Borrowed(asset, borrowAmount, interestRateMode);
+    }
+
+    /**
+     * @notice Repays a borrowed asset to Aave pool
+     * @param asset Address of the asset to repay
+     * @param repayAmount Amount of asset to repay
+     * @param interestRateMode Interest rate mode: 1 for Stable, 2 for Variable
+     */
+    function aaveRepay(
```
rndquu: I think both `aaveBorrow()` and `aaveRepay()` can be removed since we don't plan to borrow against collateral in the `UbiquityPool`.

packages/contracts/src/dollar/amo/AaveAmo.sol
```diff
@@ -0,0 +1,272 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.19;
```
rndquu: Pls use a strict version everywhere `pragma solidity 0.8.19;`

rndquu (changes requested):
Overall this PR looks good:
- natspecs everywhere
- unit tests make sense
- owner seems to able to transfer collateral to aave

There're a couple of things that we'll have to work out (not part of this PR):
- remove forked tests in favor of unit tests
- `AmoMinter` owner should be able to transfer tokens to aave in a single tx (right now owner has to [execute](https://github.com/ubiquity/ubiquity-dollar/blob/02bbe8f425bca7d720e4d0599a86eb28b25a77f4/packages/contracts/src/dollar/amo/AaveAmo.sol#L119) a 2nd tx to deposit collateral to aave)
- `IAmo` could have `deposit()` and `withdraw()` methods (which fits at least with lending and DEX protocols)

I will review this PR in detail on Monday. Meanwhile pls fix minor issues in the comments.

packages/contracts/src/dollar/amo/AaveAmo.sol
zugdev: Do you prefer to keep 2 versions of it? One with borrow and another without just in case? Could keep it configurable as well, if needed.

zugdev:
> - `AmoMinter` owner should be able to transfer tokens to aave in a single tx (right now owner has to [execute](https://github.com/ubiquity/ubiquity-dollar/blob/02bbe8f425bca7d720e4d0599a86eb28b25a77f4/packages/contracts/src/dollar/amo/AaveAmo.sol#L119) a 2nd tx to deposit collateral to aave)

There are a couple ways to do that. Adding AMO contracts to the main diamond or a specific AMO diamond is one of them. If so the IAmo concept will be extinguished, as function names can't be shared. 

> - `IAmo` could have `deposit()` and `withdraw()` methods (which fits at least with lending and DEX protocols)

Another oposing idea could be adhering to this IAmo idea, keep standalone AMO contracts and implement overriden deposit and withdraw methods so that `function giveCollateralToAmo` receives a fixed set of arguments. I am against this approach as the arguments passed in depositing/withdrawing would become static, limiting a lot possivle strategy branches - as different deposit functions require different arguments. But it is indeed an option.

zugdev:
> - `AmoMinter` owner should be able to transfer tokens to aave in a single tx (right now owner has to [execute](https://github.com/ubiquity/ubiquity-dollar/blob/02bbe8f425bca7d720e4d0599a86eb28b25a77f4/packages/contracts/src/dollar/amo/AaveAmo.sol#L119) a 2nd tx to deposit collateral to aave)

There are a couple ways to do that. Adding AMO contracts to the main diamond or a specific AMO diamond is one of them. If so the IAmo concept will be extinguished, as function names can't be shared. 

> - `IAmo` could have `deposit()` and `withdraw()` methods (which fits at least with lending and DEX protocols)

packages/contracts/src/dollar/amo/AaveAmo.sol
rndquu: What do you mean by "2 versions of it"?

zugdev: Keep fully capable (with borrow and repay) AMO contract as `AaveBorrowingAmo` under deprecated folder, and the one without borrowing as `AaveAmo` or `AaveSupplyingAmo`.  That is if borrowing against collateral, or looping/leveraging comes in handy, the contract will already be written.

zugdev: I've refactored to remove borrowing. If needed later can rollback to commit [ee11e95](https://github.com/ubiquity/ubiquity-dollar/pull/956/commits/ee11e957aab78f39fcb37192a8362527c199eb5a)

rndquu (approved):
This PR implements basic functionality for the AMO minter and aave AMO. I've opened another [issue](https://github.com/ubiquity/ubiquity-dollar/issues/968) for polishing this PR and preparing deployment migrations. 

Then we could utilize https://www.findaudit.xyz/ or ask `0xAdrii` to look though the contracts.

@0x4007 FYI

PR #957, opened 2024-09-08
Add sync-template.yml workflow that synchronizes the repo with ubiquity/ts-template every month

zyrafal:
Details: https://github.com/ubiquity/ts-template/issues/54


0x4007:
I have a feeling this might be a mess so perhaps we should not include in this repo

Issue #972, opened 2024-10-03
CI: fix `check_storage_layout` for new contracts


rndquu:
There's the [core-contracts-storage-check](https://github.com/ubiquity/ubiquity-dollar/blob/20799bedf5ee4d40d2863653f7bc2fe68a2ce3fa/.github/workflows/core-contracts-storage-check.yml) workflow which checks that contracts' storage is not messed on contract upgrades.

Check these CI runs which fails with the `No workflow run found with an artifact named...` error:
- https://github.com/ubiquity/ubiquity-dollar/actions/runs/11125075713/job/30912062971?pr=959
- https://github.com/ubiquity/ubiquity-dollar/actions/runs/11145695030/job/30975996318?pr=971

The issue is that when a **new contract** is added the [core-contracts-storage-check](https://github.com/ubiquity/ubiquity-dollar/blob/20799bedf5ee4d40d2863653f7bc2fe68a2ce3fa/.github/workflows/core-contracts-storage-check.yml) starts to fail while the expected behavior for the CI workflow to pass because we don't need to check storage layout for newly added contracts (because storage of the newly added contract hasn't been updated hence storage collision could not occur).

So as a part of this issue we should make sure that:
1. [core-contracts-storage-check](https://github.com/ubiquity/ubiquity-dollar/blob/20799bedf5ee4d40d2863653f7bc2fe68a2ce3fa/.github/workflows/core-contracts-storage-check.yml) is not failing for newly added contracts
2. [diamond-storage-check.yml](https://github.com/ubiquity/ubiquity-dollar/blob/20799bedf5ee4d40d2863653f7bc2fe68a2ce3fa/.github/workflows/diamond-storage-check.yml) is not failing for newly added libraries

P.S. Pls provide a QA in the related PR with these scenarios:
1) No storage updates, CI passing
2) Storage update, no collision, CI passing
3) Storage update, collision, CI failing
4) New contract added, CI passing

zugdev:
Can use [tj-actions/changed-files](https://github.com/tj-actions/changed-files) for this.

rndquu:
> Can use [tj-actions/changed-files](https://github.com/tj-actions/changed-files) for this.

https://github.com/ubiquity/ubiquity-dollar/blob/20799bedf5ee4d40d2863653f7bc2fe68a2ce3fa/.github/workflows/core-contracts-storage-check.yml#L25

rndquu:
[...]

PR #973, opened 2024-10-03
feat: fix core contracts storage check

zugdev:
Resolves #972


zugdev:
this looks very ugly because of rebase, i'll do better

PR #974, opened 2024-10-04
feat: skip new contracts on `check_storage_layout`

zugdev:
Resolves #972


zugdev:
@rndquu Here are the QAs, new contract is working great now, but I can't reproduce a collision at my local fork for some reason.

> 1. No storage updates, CI passing

[here](https://github.com/zugdev/ubiquity-dollar/actions/runs/11171809638)

> 2. Storage update, no collision, CI passing

[here](https://github.com/zugdev/ubiquity-dollar/actions/runs/11171648305)

> 3. Storage update, collision, CI failing

[here](https://github.com/zugdev/ubiquity-dollar/actions/runs/11171665399) for some reason `check_storage_layout` is passing but it's not something I did since I haven't touched this job, only the `provide_contracts` one

> 4. New contract added, CI passing

[here](https://github.com/zugdev/ubiquity-dollar/actions/runs/11171547219/job/31056469702)

.github/workflows/core-contracts-storage-check.yml
```diff
@@ -28,6 +29,22 @@ jobs:
...
 
+      - name: Remove newly added contracts from the list
+        id: filter-new-contracts
+        run: |
+          touch contracts.txt
+          CHANGED_CONTRACTS="${{ steps.changed-contracts.outputs.contracts_all_changed_files }}"
+          
+          # Iterate through changed contracts and check if they existed in the previous commit
+          for CONTRACT in $CHANGED_CONTRACTS; do
+            if git show HEAD~1:$CONTRACT > /dev/null 2>&1; then
```
rndquu: Check this example for a PR branch:
1. Commit 1: docs updated
2. Commit 2: new contract `MyContract` is added (ci checked that `MyContract` didn't exist in the previous commit and skipped the storage check)
3. Commit 3: docs updated again (ci checked that `MyContract` existed in the previous commit and run storage check while the expected behavior is to skip it)

Here on step 3 ci will be still failing because there's no such contract in the default `development` branch.

zugdev: I don't get it.

> Commit 3: docs updated again (ci checked that MyContract existed in the previous commit and run storage check while the expected behavior is to skip it)

I think it would be skipped since contract wasn't changed and we only look at changed contracts.

```
CHANGED_CONTRACTS="${{ steps.changed-contracts.outputs.contracts_all_changed_files }}"
```

Perhaps the analogy would work for a comment? I'll test it once again.

Do you want me to only check if contract exists in `ubiquity/ubiquity-dollar` at branch `development`?

zugdev:
@rndquu you were right. Here is the QA:

[1. Should skip new contract](https://github.com/ubiquity/ubiquity-dollar/pull/974/commits/7bb1c7bc1d009ec6fc62c94d48b6fa06034d070a)

[2. Should skip new contract even if changed in a later commit](https://github.com/ubiquity/ubiquity-dollar/pull/974/commits/a6477bcd438a22aaef6c867dba92da226b7f51d8)

[3. Should fail if core contract storage is changed with collision](https://github.com/ubiquity/ubiquity-dollar/pull/974/commits/6513217889f0c4fc667f3b7619ef0062bcc5e983)

[3. Should pass if core contract storage is changed without collision](https://github.com/ubiquity/ubiquity-dollar/pull/974/commits/f344231efe85c5212ff64c3bc7424829213a17d7)

[4. Should pass if core contract is changed but storage is not updated](https://github.com/ubiquity/ubiquity-dollar/pull/974/commits/6f1f50f394a06bf8081f96659814d7dadfb4e035)

zugdev:
If you need a quick way to see workflow runs per commit you can use GitHub's [CLI](https://github.com/cli/cli) with a version younger than a year, since this change was included in [this PR.](https://github.com/cli/cli/pull/8350)

`gh run list -c 6f1f50f394a06bf8081f96659814d7dadfb4e035`

rndquu:
@zugdev I've missed this in the original issue description, could you apply the same changes to the [diamond-storage-check.yml](https://github.com/ubiquity/ubiquity-dollar/blob/20799bedf5ee4d40d2863653f7bc2fe68a2ce3fa/.github/workflows/diamond-storage-check.yml) workflow which checks that storage was not modified for Diamond facet libraries? I've increased a time estimate in the original issue to `<1 Day`.

Pls mark this PR as "ready for review" when you're finished.

zugdev:
I can't QA properly because all artifacts from `development` have expired. `Error: Error: No workflow run found with an artifact named "development.packages_contracts_src_dollar_libraries_LibAppStorage.sol-LibAppStorage.json"`. We could either: 

1. Run `core-contracts-storage-check.yml` and `diamond-storage-check.yml` with all contracts as parameters periodically in development. Writing a new action that can be manually triggered to do that.

2. Fork the [storage check repo](https://github.com/Rubilmax/foundry-storage-check/) to change how it handles not finding artifacts.

0x4007:
Adding workflow dispatch seems like a simple and good idea. Its just a line of code.

zugdev:
> Adding workflow dispatch seems like a simple and good idea. Its just a line of code.

Won't really work since `core-contracts-storage-check.yml` and `diamond-storage-check.yml` filter by changed files only. I've added [`generate-storage-artifacts.yml`](https://github.com/ubiquity/ubiquity-dollar/pull/974/files#diff-25cc8f4e900a809fdb7b3d64e07e9f6ce44686902f833e1bc2d9136d65cf323c) which will generate artifacts for core and diamond. Please check a run of it in my own repo [here](https://github.com/zugdev/ubiquity-dollar/actions/runs/11390255568). Then can can one of you please add it to development branch and run it manually? 

@0x4007 @rndquu

0x4007:
> > Adding workflow dispatch seems like a simple and good idea. Its just a line of code.
> 
> Won't really work since `core-contracts-storage-check.yml` and `diamond-storage-check.yml` filter by changed files only. I've added [`generate-storage-artifacts.yml`](https://github.com/ubiquity/ubiquity-dollar/pull/974/files#diff-25cc8f4e900a809fdb7b3d64e07e9f6ce44686902f833e1bc2d9136d65cf323c) which will generate artifacts for core and diamond. Please check a run of it in my own repo [here](https://github.com/zugdev/ubiquity-dollar/actions/runs/11390255568). Then can can one of you please add it to development branch and run it manually?
> 
> @0x4007 @rndquu

Is it easier to just merge and then run? You could always add a `workflow_dispatch` so we can manually execute from the Actions UI. [Example](https://github.com/0x4007/sync-configs-agent/pull/4/files#diff-ed77a792963208f45fe7a1d438338061c9749ce3a2df17ba4b7fde77e113157eR4).

.github/workflows/core-contracts-storage-check.yml
```diff
@@ -4,16 +4,18 @@ on:
   push:
     branches:
       - development
+      - fix-core-contracts-storage-check
```
rndquu: Seems to be a test branch, pls remove

.github/workflows/core-contracts-storage-check.yml
```diff
@@ -4,16 +4,18 @@ on:
...
 jobs:
   provide_contracts:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: recursive
+          fetch-depth: 2  # Ensure we can access the previous commit
```
rndquu: Do we still need to fetch 2 latest commits since we compare PR with the development branch?

.github/workflows/diamond-storage-check.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - development
+      - fix-core-contracts-storage-check
```
rndquu: Pls remove

.github/workflows/generate-storage-artifacts.yml
```diff
@@ -0,0 +1,64 @@
+name: Generate Storage Artifacts
+
+on:
+  workflow_dispatch
```
rndquu: Awesome, perhaps also add a cron dispatch handler?

packages/contracts/src/dollar/core/ERC20Ubiquity.sol
```diff
@@ -25,6 +25,8 @@ abstract contract ERC20Ubiquity is
     ERC20PermitUpgradeable,
     ERC20PausableUpgradeable
 {
+    // this is a non-colliding change
```
rndquu: Pls remove all updates in the contracts

rndquu (changes requested):
Overall looks good, pls address the comments, once it's done I'll properly QA both for core and diamond lib contracts.

.github/workflows/generate-storage-artifacts.yml
zugdev: I think artifacts can last for months, not sure here.

.github/workflows/core-contracts-storage-check.yml
zugdev: I don't think so

zugdev:
waiting on PR review

0x4007:
@rndquu

rndquu (changes requested):
@zugdev Hey, sorry for a such a long review.

Check [this](https://github.com/rndquu/ubiquity-dollar/pull/38/commits/154e4a9a0d115aab57a6d6c184557e53f6b819ce) commit which [adds](https://github.com/rndquu/ubiquity-dollar/pull/38/commits/154e4a9a0d115aab57a6d6c184557e53f6b819ce#diff-2970b3bcdfa5558d89da6127812924755df903dc8330685fa6acfc91a011ee36R44) a new property at the beginning of a struct. The [diamond-storage-check](https://github.com/rndquu/ubiquity-dollar/blob/94aa536ae0693bb8439cafde8eda3a953dc2a85f/.github/workflows/diamond-storage-check.yml) workflow should fail for this case but it [passes](https://github.com/rndquu/ubiquity-dollar/actions/runs/13013154995/job/36295677604) unexpectedly.

.github/workflows/diamond-storage-check.yml
```diff
@@ -81,8 +81,8 @@ jobs:
         with:
           version: nightly
         
-      - name: Check For Diamond Storage Changes
-        uses: ubiquity/foundry-storage-check@main
+      - name: Check For Core Contracts Storage Changes
+        uses: Rubilmax/foundry-storage-check@main
```
0x4007: Probably more appropriate to update our fork.

zugdev: Actually the problem is `forge inspect <CONTRACT_NAME> storageLayout`, which is used by Rubilmax returns null for libraries, so our fork manually creates a layout storage for the libraries. Apparently it might be broken. Therefore, we have to fix our fork, using that action won't really work.

0x4007: > we have to fix our fork

Yes that's what I meant

PR #981, opened 2024-12-17
feat: support Drosera

### LINKED ISSUE #979: Drosera integration
We should add https://www.drosera.io testnet support.

What should be done:
1. [Deploy](https://github.com/ubiquity/ubiquity-dollar/blob/6ecefcad4145c360d95589d6a3dffbf1b2e08781/packages/contracts/package.json#L31) ubiquity protocol contracts to the testnet
2. Create a [trap](https://dev.drosera.io/docs/trappers/creating-a-trap) which should pause protocol contracts in case there's a security incident
3. Make sure the trap actually works (i.e. pauses ubiquity contracts) by triggering conditions leading to trap execution

kingsley-einstein:
Resolves [#979](https://github.com/ubiquity/ubiquity-dollar/issues/979)


.gitmodules
```diff
@@ -46,3 +46,6 @@
 	path = packages/contracts/lib/aave-v3-periphery
 	url = https://github.com/aave/aave-v3-periphery
 	branch= master
+[submodule "packages/contracts/lib/v3-periphery"]
```
rndquu, line 49: Why do we need this submodule? It's not used anywhere.

packages/contracts/src/traps/ProtocolTrap.sol
```diff
@@ -0,0 +1,35 @@
+pragma solidity ^0.8.0;
+
+contract ProtocolTrap {
```
rndquu, line 3: Contract must implement `ITrap`.

packages/contracts/src/traps/ProtocolTrap.sol
```diff
@@ -0,0 +1,35 @@
...
+    struct CollectOutput {
+        bool triggerResponse;
+    }
+
+    function collect() external view returns (bytes memory) {
+        uint256 randomNumber = randomize();
```
rndquu, line 9: What's the point of returning a random number?

rndquu (changes requested):
@kingsley-einstein 

1. Pls fix the comments
2. Did you deploy the contracts to the testnet?
3. Did you check that the trap works as expected?
4. The trap must pause the contracts.
5. CI should pass.

kingsley-einstein:
@rndquu 
The trap contract has been deployed. For it to work, I think an operator has to opt into it. I'm guessing I'll have to opt to become an operator as no operator has opted in since I deployed the contract.

rndquu:
> @rndquu The trap contract has been deployed. For it to work, I think an operator has to opt into it. I'm guessing I'll have to opt to become an operator as no operator has opted in since I deployed the contract.

Drosera runs its own operators and drosera team made us sure there will be at least 1 operator in our trap. So becoming an operator is unnecessary, it's drosera's job.

PR #987, opened 2025-03-25
feat: Staking v2

### LINKED ISSUE #985: Staking v2
We have the [old staking contract](https://github.com/ubiquity/ubiquity-dollar/wiki/24.-Smart-Contracts#masterchef-v21) (which still works) where users are able to stake 3CRV LP tokens and get rewards in UBQ tokens. The old staking contract has some issues:
1. It uses 3CRV LP tokens from the old `uAD/3CRV` pool. We've redeployed our stable coin so now we're using [UUSD](https://github.com/ubiquity/ubiquity-dollar/wiki/24.-Smart-Contracts#ubiquitydollartoken) instead of the deprecated [uAD](https://github.com/ubiquity/ubiquity-dollar/wiki/08.-uAD) token.
2. The old staking contract is unaudited
3. There's [this](https://github.com/ubiquity/ubiquity-dollar/issues/939) issue with the old staking contract

It makes sense to reimplement staking so that users were able to stake `UUSD/LUSD` [LP tokens](https://github.com/ubiquity/ubiquity-dollar/wiki/24.-Smart-Contracts#lusduusd) and get `UBQ` rewards. This way contributors have an incentive to keep `UUSD` in the `UUSD/LUSD` pool instead of swapping `UUSD` for a more liquid asset.

We've already ported staking to [this](https://github.com/ubiquity/ubiquity-dollar/blob/6f3b3e02dcc342bc39f7f4cae3112a51cbfc8d58/packages/contracts/src/dollar/facets/StakingFacet.sol) diamond facet but we've never tested how it really works.

What should be done:
1. Refactor [LibStaking](https://github.com/ubiquity/ubiquity-dollar/blob/6f3b3e02dcc342bc39f7f4cae3112a51cbfc8d58/packages/contracts/src/dollar/libraries/LibStaking.sol) to match the original [MaterChef](https://github.com/sushi-labs/sushiswap/blob/271458b558afa6fdfd3e46b8eef5ee6618b60f9d/contracts/MasterChef.sol) implementation
2. Cover `LibStaking` with unit tests
3. Prepare on-chain [migrations](https://github.com/ubiquity/ubiquity-dollar/tree/development/packages/contracts/migrations)

The next step would be to audit the updated code and deploy to production.

rndquu:
Resolves https://github.com/ubiquity/ubiquity-dollar/issues/985

This PR:
1. Adds a feature of [staking](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/src/dollar/libraries/LibStaking.sol) (based on [MasterChef](https://github.com/sushi-labs/sushiswap/blob/271458b558afa6fdfd3e46b8eef5ee6618b60f9d/contracts/MasterChef.sol) with slight changes) where users are able to stake `LUSD/UUSD LP` tokens from Curve mainnet in order to get `UBQ` rewards
2. Removes old staking related contracts
3. Adds a mainnet [migration](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/migrations/mainnet/Deploy002_Staking.s.sol) for staking deployment

### How to QA

Unit [tests](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/test/diamond/facets/StakingFacet.t.sol) pretty much QA the feature, in particular [staking](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/test/diamond/facets/StakingFacet.t.sol#L207-L241) and [unstaking](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/test/diamond/facets/StakingFacet.t.sol#L249-L283)

### How to QA staking mainnet migration

There is the staking [migration](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/migrations/mainnet/Deploy002_Staking.s.sol) script which deploys [StakingFacet](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/src/dollar/facets/StakingFacet.sol) to `mainnet`. You may QA it this way:
1. Run `yarn start:anvil` in one terminal window (this will run a local anvil instance forked from `mainnet`)
2. Run `yarn deploy:mainnet:qa` in another terminal window (this will impersonate `ubq.eth` address and deploy [StakingFacet](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/src/dollar/facets/StakingFacet.sol))
3. Add the [these](https://gist.github.com/rndquu/046f86ebb0eb32315c62f66db38bfc40#file-test-sol-L11-L13) imports and [this](https://gist.github.com/rndquu/046f86ebb0eb32315c62f66db38bfc40#file-test-sol-L16-L57) test to [StakingFacet.t.sol](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/test/diamond/facets/StakingFacet.t.sol)
4. From `packages/contracts` folder run `forge test --match-test testStakingMigrationMainnet` which will check that staking and unstaking work as expected for newly deployed [StakingFacet](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/src/dollar/facets/StakingFacet.sol)

NOTICE: 
1. The `Check For Diamond Storage Changes` CI workflow is [failing](https://github.com/ubiquity/ubiquity-dollar/actions/runs/14336021531/job/40183282004?pr=987) with the `The artifact name is not valid` error. I'm not really sure about the root cause but even if that error was not present the `Check For Diamond Storage Changes` workflow would still be failing because `LibStaking` [storage](https://github.com/ubiquity/ubiquity-dollar/blob/ac517749fdddef3b9755c305224e2aacf14b7756/packages/contracts/src/dollar/libraries/LibStaking.sol#L57-L78) is totally updated. Anyway it's safe to update `LibStaking` since the `StakingFacet` is not yet deployed to production. The `The artifact name is not valid` error will we addressed in a separate github [issue](https://github.com/ubiquity/ubiquity-dollar/issues/992).
2. The `Compare Test Coverage` CI workflow started to [fail](https://github.com/ubiquity/ubiquity-dollar/actions/runs/14779246508/job/41494323234?pr=987) with the `FAIL: 51` error for the AMO minter related tests. It happened because we're using forked tests for the AMO minter feature and it seems that aave pool on ethereum sepolia reached it supply capacity hence the `SUPPLY_CAP_EXCEEDED = 51` error is thrown (more [info](https://mirror.xyz/unfrigginbelievable.eth/fzvIBwJZQKOP4sNpkrVZGOJEk5cDr6tarimQHTw6C84)). Anyway current PR fixes that CI error in [this](https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/05fe2f73c6bc20bebd273505a18c60619f7fe787) commit. The `development` branch doesn't have the fix yet hence this [fail](https://github.com/ubiquity/ubiquity-dollar/actions/runs/14779246508/job/41494323234?pr=987) happens on getting test coverage for the `development` branch step. Once this PR is merged the error will be gone.

The next step is to audit the staking implementation and deploy to production.


rndquu:
Updated UBQ token emissions to 0.2 UBQ per block according to [this](https://github.com/ubiquity/ubiquity-dollar/issues/985#issuecomment-2796896966) comment

zugdev:
hey, I've checked the math and general implementation looks fine. also seems very modular and well adapted for a diamond pattern, good job. I have yet to verify the tests to ensure you've covered all scenarios. lmk if there is a specific direction you are unsure about, and want me to explore on

rndquu:
> lmk if there is a specific direction you are unsure about, and want me to explore on

Nothing special to mention, our staking feature is basically a battle-tested [MasterChef](https://github.com/sushi-labs/sushiswap/blob/271458b558afa6fdfd3e46b8eef5ee6618b60f9d/contracts/MasterChef.sol)

zugdev:
looks good

zugdev (approved):

0x4007:
@EresDev @whilefoo any thoughts on this pull? I don't know Solidity well enough to review it.

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,451 @@
...
+        if (lpSupply == 0) {
+            pool.lastRewardBlock = block.number;
+            return;
+        }
+        uint256 multiplier = getStakingMultiplier(pool.lastRewardBlock, block.number);
+        uint256 governanceReward =
+            multiplier.mul(stakingStore.governancePerBlock).mul(pool.allocationPoints).div(
+                stakingStore.totalAllocationPoints
+            );
+        stakingStore.rewardToken.mint(store.treasuryAddress, governanceReward.div(stakingStore.governanceTreasuryDivider));
```
Zgorizzo: note that the inflation of the gov token will not only be dictated by the  governancePerBlock but by governancePerBlock + 100/governanceTreasuryDivider %

packages/contracts/test/diamond/facets/StakingFacet.t.sol
```diff
@@ -1,376 +1,502 @@
...
+        assertEq(poolInfo.lastRewardBlock, 11);
     }
 
-    function testSetBlockCountInAWeek(uint256 x) public {
-        vm.expectEmit(true, false, false, true);
-        emit BlockCountInAWeekUpdated(x);
-        vm.prank(admin);
-        stakingFacet.setBlockCountInAWeek(x);
-        assertEq(x, stakingFacet.blockCountInAWeek());
+    function testStake_ShouldStakeTokens() public {
```
Zgorizzo, line 259: I would add fuzzing for stake and unstake
+ add staking unstaking test with multiple users 
+ add test staking with an increase/decrease of GovernancePerBlock

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,451 @@
...
-            // and remove the lp Reward Debt
-            return
-                (bs[0] * (curAccLpRewardPerShare)) /
-                (1e12) -
-                (stake.lpRewardDebt);
-        }
-        return 0;
+    function setGovernanceBonusEndBlock(uint256 newGovernanceBonusEndBlock) internal {
+        StakingStorage storage stakingStore = stakingStorage();
+        stakingStore.bonusEndBlock = newGovernanceBonusEndBlock;
```
Zgorizzo, line 442: need some upper lower bound  checks

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,451 @@
...
-    ) internal view returns (uint256 pendingLpReward) {
-        StakingData storage ss = stakingStorage();
-        if (ss.accLpRewardPerShare > 0) {
-            pendingLpReward =
-                (amount * ss.accLpRewardPerShare) /
-                1e12 -
-                (lpRewardDebt);
-        }
+    function setGovernanceBonusMultiplier(uint256 newGovernanceBonusMultiplier) internal {
+        StakingStorage storage stakingStore = stakingStorage();
```
Zgorizzo, line 453: need some checks

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,451 @@
...
-            .appStorage()
-            .stakingShareAddress;
-        // priceShare = totalLP / totalShares
-        priceShare = LibStakingFormulas.bondPrice(
-            StakingShare(stakingShareAddress).totalLP(),
-            totalShares,
-            ONE
-        );
+    function setGovernancePerBlock(uint256 newGovernancePerBlock) internal {
+        StakingStorage storage stakingStore = stakingStorage();
```
Zgorizzo, line 464: all set function need input param checks

Zgorizzo (commented):
reward calculation is iso with masterchef
although it needs more tests IMHO especially fuzzing around staking/unstaking with more than one user or with changing rewards (gov and for treasury)
and more input params checks

0x4007:
Thanks a ton @Zgorizzo !

@rndquu, he wrote the original Ubiquity Dollar including the old staking.

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: > note that the inflation of the gov token will not only be dictated by the governancePerBlock but by governancePerBlock + 100/governanceTreasuryDivider %

Yes, it works exactly as the current staking implementation

Example params:
- `governancePerBlock: 1e18`
- `governanceTreasuryDivider: 5`

For the parameters above `1.2 UBQ` will be minted each block: `1 UBQ` for staking rewards + `1 UBQ / governanceTreasuryDivider = 1 / 5 = 0.2 UBQ` to the treasury

packages/contracts/test/diamond/facets/StakingFacet.t.sol
rndquu: Added in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/2329f351eda07ff332243088cff14a314ef6e419

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: Added in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/b1496707e319d6417f37103040edfa578621f7a4

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: Added in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/b1496707e319d6417f37103040edfa578621f7a4

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: Added in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/b1496707e319d6417f37103040edfa578621f7a4

rndquu:
@Zgorizzo All comments are addressed:
1. Added more validation checks (where it makes sense)
2. Added a fuzzing test

Ready for review

NOTICE: The Compare Test Coverage CI workflow started to [fail](https://github.com/ubiquity/ubiquity-dollar/actions/runs/14779246508/job/41494323234?pr=987) with the `FAIL: 51` error for the AMO minter related tests. It happened because we're using forked tests for the AMO minter feature and it seems that aave pool on ethereum sepolia reached it supply capacity hence the `SUPPLY_CAP_EXCEEDED = 51` error is thrown (more [info](https://mirror.xyz/unfrigginbelievable.eth/fzvIBwJZQKOP4sNpkrVZGOJEk5cDr6tarimQHTw6C84)). Anyway current PR fixes that CI error in [this](https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/05fe2f73c6bc20bebd273505a18c60619f7fe787) commit. The development branch doesn't have the fix yet hence this [fail](https://github.com/ubiquity/ubiquity-dollar/actions/runs/14779246508/job/41494323234?pr=987) happens on getting test coverage for the development branch step. Once this PR is merged the error will be gone.

whilefoo:
What is the blocker here?

ryzhak:
> What is the blocker here?

Lack of 2nd approve of the PR, PR is ready for review

0x4007:
@Whilefoo maybe you can take a look into this? I also asked zgo one more time.

whilefoo:
I can make a review (even though I'm not really an expert in Solidity)

0x4007:
Perhaps it makes sense to reach out to some of the auditors at Sherlock as well then.

packages/contracts/test/fuzz/diamond/facets/StakingFacet.fuzz.t.sol
```diff
@@ -0,0 +1,150 @@
...
+        vm.prank(user);
+        stakingFacet.stake(0, stakeAmount);
+        vm.prank(user2);
+        stakingFacet.stake(0, stakeAmount);
+
+        // Y blocks pass
+        vm.roll(block.number + blocksPassed);
+
+        // users collect rewards
+        vm.prank(user);
```
zgorizzo69, line 216: would be  nice to have staking at different time to make sure the calculation is ok

ryzhak: I don't fully understand the "at different time" statement.

Staking rewards depend on number of blocks passed, not on a `block.timestamp`. There's already a [time gap](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/test/fuzz/diamond/facets/StakingFacet.fuzz.t.sol#L103) between staking and collecting rewards which basically uses [arbitrary](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/test/fuzz/diamond/facets/StakingFacet.fuzz.t.sol#L62) number of blocks passed so not really sure how to improve the test.

zgorizzo69: you can have userA deposits amountA at  time T and userB deposits amountB  at time T+x and verify at time T' (where T' > T+x) that when UserA witdraw he gets the right amount and then at T'' (where T''>T') UserB withdraw and gets the right amount.  here we have UserA and UserB deposit the same amount at the same time and withdraw at the same time

packages/contracts/src/dollar/libraries/LibStaking.sol
zgorizzo69: I would also add upper bound limit (I don't think we would allow 10000 ether per block reward for instance and add the corresponding fuzzing test i.e setGovernancePerBlock should work with a fuzzing newGovPerBlock between 0 and UPPER_BOUND

packages/contracts/src/dollar/libraries/LibStaking.sol
zgorizzo69: I would also add upper bound limit + fuzzing equivalent test

packages/contracts/src/dollar/libraries/LibStaking.sol
zgorizzo69: I would also add upper bound limit to avoid setting it 10 years from now for instance+ fuzzing equivalent test

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-        // priceShare = totalLP / totalShares
-        priceShare = LibStakingFormulas.bondPrice(
-            StakingShare(stakingShareAddress).totalLP(),
-            totalShares,
-            ONE
+    function setGovernanceTreasuryDivider(
+        uint256 newGovernanceTreasuryDivider
+    ) internal {
+        require(
+            newGovernanceTreasuryDivider > 0,
```
zgorizzo69, line 479: I would also add upper bound limit + fuzzing equivalent test

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-                ss.accLpRewardPerShare =
-                    ss.accLpRewardPerShare +
-                    (((currentLpRewards - ss.lpRewards) * 1e12) / totalShares);
-
-                // update the staking contract lpRewards
-                ss.lpRewards = currentLpRewards;
-            }
-        }
+    function setStakingRewardToken(address newRewardToken) internal {
+        require(newRewardToken != address(0), "Zero address detected");
```
zgorizzo69, line 492: btw isn't it the time to use `error ZeroAddress()` instead of string errors ?

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-        return
-            StakingShare(LibAppStorage.appStorage().stakingShareAddress).mint(
-                to,
-                lpAmount,
-                lpRewardDebt,
-                endBlock
-            );
+    function setStakingStartBlock(uint256 newStartBlock) internal {
+        require(newStartBlock >= block.number, "Can't start in the past");
+        StakingStorage storage stakingStore = stakingStorage();
```
zgorizzo69, line 504: I would emit events everywhere we modify the state for easy integration and improved UX

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
+        uint256 poolId,
+        uint256 allocationPoints,
+        bool withUpdate
+    ) internal {
+        StakingStorage storage stakingStore = stakingStorage();
+
+        require(poolId < stakingStore.poolInfo.length, "Pool does not exist");
+
+        if (withUpdate) {
+            massUpdateStakingPools();
```
zgorizzo69: I would add a `poolLength` as a parameter instead of a `bool` so that if `massUpdateStakingpools` reverts due to all gas being consumed we can still call it with a smaller poolLength

packages/contracts/test/diamond/facets/StakingFacet.t.sol
```diff
@@ -1,376 +1,630 @@
...
+        );
+        assertEq(poolInfo.lastRewardBlock, 1);
+
+        // 10 blocks pass
+        vm.roll(block.number + 10);
+
+        stakingFacet.massUpdateStakingPools();
+
+        // after
+        poolInfo = stakingFacet.getStakingPoolInfo(0);
```
zgorizzo69, line 253: we need to test this func with more pools if we want it to be massive

packages/contracts/test/fuzz/diamond/facets/StakingFacet.fuzz.t.sol
```diff
@@ -0,0 +1,150 @@
...
+        assertEq(rewardToken.balanceOf(user), 0);
+        assertEq(rewardToken.balanceOf(user2), 0);
+        assertEq(stakeToken.balanceOf(address(stakingFacet)), 0);
+        assertEq(userInfo.amount, 0);
+        assertEq(userInfo2.amount, 0);
+        assertEq(poolInfo.amount, 0);
+
+        // users stake STK
+        vm.prank(user);
+        stakingFacet.stake(0, stakeAmount);
```
zgorizzo69, line 205: would be nice to fuzz pool creation and stake on another fuzz test

zgorizzo69 (changes requested):

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-            1e12;
-
-        StakingShare(stakingShareAddress).updateStake(
-            _id,
-            stake.lpAmount,
-            stake.lpRewardDebt,
-            stake.endBlock
+        if (withUpdate) {
+            massUpdateStakingPools();
+        }
```
whilefoo: Wouldn't it make sense to always update pools because after adding a new pool the allocation points will change?

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-        );
-        uint256 toTransfer = IERC20(metaPool.coins(1)).balanceOf(address(this));
+    function getStakingMultiplier(
+        uint256 from,
+        uint256 to
+    ) internal view returns (uint256) {
+        StakingStorage storage stakingStore = stakingStorage();
+
+        if (to <= stakingStore.bonusEndBlock) {
+            return to.sub(from).mul(stakingStore.governanceBonusMultiplier);
```
whilefoo, line 186: Since multiplier is just a uint, multiplier can only be 0 = 0%, 1 = 100%, 2 = 200%, 3 = 300%...and not 120%?

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-                (bs[0] * (curAccLpRewardPerShare)) /
-                (1e12) -
-                (stake.lpRewardDebt);
-        }
-        return 0;
+    function setGovernanceBonusMultiplier(
+        uint256 newGovernanceBonusMultiplier
+    ) internal {
+        StakingStorage storage stakingStore = stakingStorage();
+        stakingStore.governanceBonusMultiplier = newGovernanceBonusMultiplier;
```
whilefoo, line 454: I'd add a check that multiplier is higher than 0

packages/contracts/src/dollar/libraries/LibStaking.sol
whilefoo: Here I'd also always update the pools because allocation points change

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-            amount,
-            1,
-            0
-        );
-        uint256 toTransfer = IERC20(metaPool.coins(1)).balanceOf(address(this));
+    function getStakingMultiplier(
+        uint256 from,
+        uint256 to
+    ) internal view returns (uint256) {
+        StakingStorage storage stakingStore = stakingStorage();
```
whilefoo, line 183: I'd check that `from` is lower than `to`

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-            pendingLpReward
+        PoolInfo storage pool = stakingStore.poolInfo[poolId];
+        if (block.number <= pool.lastRewardBlock) {
+            return;
+        }
+        uint256 lpSupply = pool.amount;
+        if (lpSupply == 0) {
+            pool.lastRewardBlock = block.number;
+            return;
+        }
```
whilefoo, line 365: I would return early if `stakingStore.totalAllocationPoints` is 0 to prevent reverting.
For example if you have multiple pools and you set all their allocations points to 0 to pause them, then stake /unstake would fail because `updateStakingPool` would revert because of diving by 0.

I guess it would also make sense to return early if `pool.allocationPoints` is 0 to not waste gas when doing a mass update

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
-            _sharesAmount,
-            _weeks,
-            _endBlock
-        );
+    /**
+     * @notice Unstakes LP tokens from the staking contract
+     * @param poolId Pool id
+     * @param amount Amount of LP tokens to unstake
+     */
+    function unstake(uint256 poolId, uint256 amount) internal {
```
whilefoo, line 326: I guess I'd be a good idea to add a function for emergency unstaking in case something with minting or calculations go wrong so people can still withdraw their LP tokens

packages/contracts/src/dollar/libraries/LibStaking.sol
```diff
@@ -1,532 +1,514 @@
...
+        uint256 rewardDebt; // Reward debt. See explanation below.
+    }
+
+    /// @notice Info of each pool
+    struct PoolInfo {
+        IERC20 lpToken; // Address of LP token contract.
+        uint256 amount; // Total amount of LP tokens staked in a pool.
+        uint256 allocationPoints; // How many allocation points assigned to this pool. Governance tokens to distribute per block.
+        uint256 lastRewardBlock; // Last block number that Governance tokens distribution occurs.
+        uint256 accumulatedGovernancePerShare; // Accumulated Governance tokens per share, times 1e12. See below.
```
whilefoo, line 52: why does it have to be multiplied by 1e12?

packages/contracts/test/diamond/facets/StakingFacet.t.sol
```diff
@@ -1,376 +1,630 @@
...
-
-    event MinPriceDiffToUpdateMultiplierModified(
-        uint256 indexed minPriceDiffToUpdateMultiplier
-    );
-    event BlockCountInAWeekUpdated(uint256 _blockCountInAWeek);
-    event StakingDiscountMultiplierUpdated(uint256 _stakingDiscountMultiplier);
-
-    function setUp() public virtual override {
+
+contract StakingFacetTest is DiamondTestSetup {
```
whilefoo, line 11: I would also test multiple pools with different allocation points and maybe one pool with 0 allocation points and also with multiple users staking in different pools

packages/contracts/test/diamond/facets/StakingFacet.t.sol
```diff
@@ -1,376 +1,630 @@
...
+
+        LibStaking.PoolInfo memory poolInfo = stakingFacet.getStakingPoolInfo(
+            0
         );
-        managerFacet.setCreditTokenAddress(address(creditToken));
+        assertEq(address(poolInfo.lpToken), address(stakeToken));
+        assertEq(poolInfo.amount, 50 ether);
+        assertEq(poolInfo.allocationPoints, 100);
+        assertEq(poolInfo.lastRewardBlock, 11);
+        assertEq(poolInfo.accumulatedGovernancePerShare, 0.0000002 ether);
```
whilefoo, line 213: also check if treasury got the correct amount of tokens

packages/contracts/test/diamond/facets/StakingFacet.t.sol
```diff
@@ -1,376 +1,630 @@
...
-        );
-        vm.startPrank(admin);
-        fourthID = stakingShare.totalSupply() + 1;
-        vm.stopPrank();
-        vm.startPrank(fourthAccount);
-        metapool.approve(address(diamond), fourthBal);
-        stakingFacet.deposit(fourthBal, 1);
+        (, , uint256 newGovernanceBonusMultiplier, , , , , ) = stakingFacet
+            .getStakingSettings();
+        assertEq(newGovernanceBonusMultiplier, 10);
```
whilefoo: I don't know if it makes sense but I'd test if changing parameters like bonus multiplier, bonus end block, starting block, reward token, governance per block... don't break the calculations.

For example if you change governance per block then the new value should be applied from the last reward block forward, or if you change start block when pools already have funds.

packages/contracts/test/diamond/facets/StakingFacet.t.sol
rndquu: [This](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/test/diamond/facets/StakingFacet.t.sol#L173-L192) test is checking that [getStakingPoolInfo](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L227-L232) returns expected pool parameters. It doesn't track the amount of tokens sent to the treasury.

The test that checks that treasury receives expected amount of tokens already [exists](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/test/diamond/facets/StakingFacet.t.sol#L349-L382).

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: [massUpdateStakingPools](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L250) is basically a helper method so that when admin [adds](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L369) a new pool or [updates](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L477) an existing one all pool parameters will be fresh. [massUpdateStakingPools](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L250) can be safely removed from the codebase because:
- users update pool parameters on [staking](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L269) and [unstaking](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L303) 
- any user can [update](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L324) a single pool

I suppose the main reason why `SushiSwap` ever introduced the [massUpdateStakingPools](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L250) method is to keep their UI consistent when new pools are added/edited.

The main reason why `massUpdateStakingPools` is wrapped in `if (withUpdate)` [here](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L378) and [here](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L486) is that if there're too many pools that consume all available gas in a block on mass pool update then it won't be possible to "mass update" pools on adding/editing a pool (though it would be still possible to [update](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L324) them one by one). 

Even if we introduce some kind of range of pool ids for "mass update" then [here](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L379) and [here](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L487) UI would still be inconsistent (because not all pools are updated).

To sum up there're at least 3 options:
1. Keep it as is. Right now a single [updateStakingPool](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L324) consumes 131k of gas. Given that ethereum mainnet has a 45mln block gas limit we can freely "mass update" 343 staking pools. Even if we ever reach that limit then it doesn't affect the staking rewards distribution or whatever related to rewards at all. 
2. Remove `massUpdateStakingPools` at all. 
3. Remove [these](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L378-L380) and [these](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L486-L488) lines and introduce a range of pool ids for update in [massUpdateStakingPools](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L250). Somebody (admin?) will have to call it from time to time in order to keep the UI consistent for all pools (not just the ones that are frequently used).

> Wouldn't it make sense to always update pools because after adding a new pool the allocation points will change?

No, because it may run out of gas if there're too many pools. In the end users use fresh data because pools are updated on [staking](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L269) and [unstaking](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L303).

whilefoo: My reasoning was that if you add a pool but don't update the pools then the next time it will do an update, the rewards will be calculated on the new allocation points and lastRewardBlock will be  older than the block when the pool was actually added. I guess it would only make a difference if the pool is not used a lot and the last reward block is old.

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: Yes, as a side note we don't [plan](https://github.com/ubiquity/ubiquity-dollar/issues/985#issuecomment-2776348113) to use that feature at all

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: I've double checked, the [testUnstake_ShouldUnstakeTokens](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/test/diamond/facets/StakingFacet.t.sol#L271) passes (i.e. user receives the same amount of rewards) both with and without [mass updating](https://github.com/ubiquity/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/test/diamond/facets/StakingFacet.t.sol#L50) pools on pool creation (if that's what you meant).

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: We don't [plan](https://github.com/ubiquity/ubiquity-dollar/issues/985#issuecomment-2776348113) (question 6) to use this feature at all so it will be 0 in production deployment.

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: Refer to [this](https://github.com/ubiquity/ubiquity-dollar/pull/987#discussion_r2253552899) comment

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: There's a little difference whether the method reverts with an underflow or via a `require` statement, in the end the whole tx reverts

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: In order to reduce precision loss errors, everywhere in the code where `accumulatedGovernancePerShare` is used ([one](https://github.com/rndquu/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L274), [two](https://github.com/rndquu/ubiquity-dollar/blob/2329f351eda07ff332243088cff14a314ef6e419/packages/contracts/src/dollar/libraries/LibStaking.sol#L307), etc...) the value is then divided by `1e12` common denominator

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: I don't really want to introduce more methods for withdrawing any kind of assets because it widens the attack surface. Remember that we're using the diamond proxy pattern. It means that **all of our assets** (i.e. Ubiquity pool collateral, staked tokens, etc...) reside in a single contract. We already can't use the staked token as collateral in the Ubiquity pool (check last sentence in https://github.com/ubiquity/ubiquity-dollar/issues/994) because it skews Ubiquity pool calculations and some invariants may be broken. I'm afraid of unforeseen attack vectors where user deposits via `Facet1` (UbiquityPool?) and somehow withdraws more that expected via `Facet2` (Staking?). 

That's why the concept is to restrict users as much as possible. In case of emergency the contracts are upgradeable.

packages/contracts/test/diamond/facets/StakingFacet.t.sol
rndquu: Added in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/1e6b295d5b7f5cf50f3231b61a382052710732c5

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: > I would return early if stakingStore.totalAllocationPoints is 0 to prevent reverting.
For example if you have multiple pools and you set all their allocations points to 0 to pause them, then stake /unstake would fail because updateStakingPool would revert because of diving by 0.

Yes, that's a valid case, I've double checked and it indeed reverts when `totalAllocationPoints == 0`. The thing is I don't want to diverge from the original [MasterChef](https://github.com/sushi-labs/sushiswap/blob/271458b558afa6fdfd3e46b8eef5ee6618b60f9d/contracts/MasterChef.sol) even with a single line of code because the original `MasterChef` hasn't yet been hacked while adding oprimizations (like early reverts) may introduce new edge cases.

>  I guess it would also make sense to return early if `pool.allocationPoints` is 0 to not waste gas when doing a mass update

That also seems a valid point but when `pool.allocationPoints == 0` the last reward block is still [updated](https://github.com/ubiquity/ubiquity-dollar/blob/1e6b295d5b7f5cf50f3231b61a382052710732c5/packages/contracts/src/dollar/libraries/LibStaking.sol#L353) and I'm not 100% sure how it affects calculations, so again, I would stick to the original [MasterChef](https://github.com/sushi-labs/sushiswap/blob/271458b558afa6fdfd3e46b8eef5ee6618b60f9d/contracts/MasterChef.sol) implementation.

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: Added in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/1422df91ad35edcf90eb32a9e127df451c5e147f

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: The whole codebase uses `require` statements for error handling so it makes sense to be consistent

packages/contracts/test/diamond/facets/StakingFacet.t.sol
rndquu: Added in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/c57b8f0ad9e2f380cdb9d7378f54424f3cd2e0a9

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: Refactored to use individual pool ids for mass update in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/c57b8f0ad9e2f380cdb9d7378f54424f3cd2e0a9

packages/contracts/src/dollar/libraries/LibStaking.sol
whilefoo: That makes sense, I was looking at a different MasterChef contract that didn't have `withUpdate`. Then `withUpdate` is even necessary because if you'd set all pools to 0, you wouldn't be able to change the allocation points.

packages/contracts/test/diamond/facets/StakingFacet.t.sol
rndquu: > For example if you change governance per block then the new value should be applied from the last reward block forward, or if you change start block when pools already have funds.

This is more of an invariant test which should be checked in https://github.com/ubiquity/ubiquity-dollar/issues/994

> I don't know if it makes sense but I'd test if changing parameters like bonus multiplier, bonus end block, starting block, reward token, governance per block... don't break the calculations.

Added more fuzzing tests in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/9c7fef6dab4d04cebffabd43593e978a7a5816ce

packages/contracts/test/fuzz/diamond/facets/StakingFacet.fuzz.t.sol
rndquu: Added more fuzzing tests in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/9c7fef6dab4d04cebffabd43593e978a7a5816ce

packages/contracts/test/fuzz/diamond/facets/StakingFacet.fuzz.t.sol
rndquu: Added more fuzzing tests with different staking/unstaking time in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/9c7fef6dab4d04cebffabd43593e978a7a5816ce

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: I don't think it makes sense to restrict the upper limit of parameters: 
- setting bonus period to a couple of years doesn't seem too strange
- setting governance reward per block to 10000 ether also seems reasonable for some cheap tokens (`1 SHIB = $0.00001232` [source](https://coinmarketcap.com/currencies/shiba-inu/))

Contracts are upgradeable so in the end admin will be able to circumvent those boundaries.

Added more fuzzing tests in https://github.com/ubiquity/ubiquity-dollar/pull/987/commits/9c7fef6dab4d04cebffabd43593e978a7a5816ce.

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: See https://github.com/ubiquity/ubiquity-dollar/pull/987#discussion_r2257853516

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: See https://github.com/ubiquity/ubiquity-dollar/pull/987#discussion_r2257853516

packages/contracts/src/dollar/libraries/LibStaking.sol
rndquu: See https://github.com/ubiquity/ubiquity-dollar/pull/987#discussion_r2257853516

rndquu:
@zgorizzo69 @whilefoo 

Ready for review, pls notice that there will be an [audit](https://github.com/ubiquity/ubiquity-dollar/issues/994) of the staking contracts which will deeply check staking invariants.

0x4007:
I think after Whilefoo review we can merge and rely more on the auditing at this point

whilefoo (approved):

PR #988, opened 2025-03-27
chore: use new sepolia RPC for tests

rndquu:


0x4007:
https://github.com/ubiquity/permit2-rpc-manager/tree/main/packages/permit2-rpc-client

As a heads up I deployed a stable rpc client and endpoint. Maybe it will be useful for you. 

Essentially takes chainlist but also works for non CORS enabled RPCs.

PR #999, opened 2025-08-26
Staking Audit Fixes

ryzhak:
Resolves https://github.com/ubiquity/ubiquity-dollar/issues/998

In particular applies changes from https://github.com/ubiquity/ubiquity-dollar/issues/998#issuecomment-3219114982.

| Issue  | Status |
| :------------- | :------------- |
| [H-01] Missing "mass pool update" on creating/updating a pool affects users' rewards  | [Fixed](https://github.com/ubiquity/ubiquity-dollar/pull/999/commits/a38e0a586c4c91c492310292a890772c67d2e32f)  |
| [M-01] Staking and reward tokens overlapping with LibUbiquityPool's collateral skews calculations  | [Fixed](https://github.com/ubiquity/ubiquity-dollar/pull/999/commits/68083a3595f85093296ab91e94d7d0b373cb951c) |
| [M-02] Possible reentrancy  | [Fixed](https://github.com/ubiquity/ubiquity-dollar/pull/999/commits/ccdba3e4b45c4d98b86a27860fdcca52effee6ec) |
| [M-03] setStakingRewardToken causes DoS of stake/unstake methods  | Acknowledged |
| [M-04] Staked tokens are stuck in the contract  | Acknowledged |
| [M-05] Staking DoS if UBQ_MINTER_ROLE is revoked from the Diamond contract  | Acknowledged |
| [M-06] User can get 0 rewards on high stake amounts  | [Fixed](https://github.com/ubiquity/ubiquity-dollar/pull/999/commits/0dcf7afb869fbd1ac01060128f0d7a333646c935) |
| [M-07] Dust reward tokens are stuck in the contract  | Acknowledged |
| [M-08] Some "weird" ERC20 tokens are not supported  | Acknowledged |
| [L-01] Unused imports  | [Fixed](https://github.com/ubiquity/ubiquity-dollar/pull/999/commits/f71092037993581f5db4e0da9004ae219300aea0) |
| [L-02] whenNotPaused modifier not used  | [Fixed](https://github.com/ubiquity/ubiquity-dollar/pull/999/commits/4c8617d1e324a810f9af430e0fa13217cbb00d0e) |
| [L-03] Treasury griefing  | [Fixed](https://github.com/ubiquity/ubiquity-dollar/pull/999/commits/0dcf7afb869fbd1ac01060128f0d7a333646c935) |
[...]

PR #1004, opened 2025-09-15
Fix staking migration

### LINKED ISSUE #1001: Staking deploy script outdated
I was trying to deploy the staking contract and ran into 2 issues.

## 1st problem

https://github.com/ubiquity/ubiquity-dollar/blob/ac3230f9ef5f4edab56316d3971862ac61a4ccd5/packages/contracts/migrations/mainnet/Deploy002_Staking.s.sol#L77-L81

The 3rd argument which was used to update all the pools was removed so it should not be provided.

## 2nd problem

https://github.com/ubiquity/ubiquity-dollar/blob/ac3230f9ef5f4edab56316d3971862ac61a4ccd5/packages/contracts/migrations/mainnet/Deploy002_Staking.s.sol#L76

This line fails with `Error: reverted with: Can't start in the past`. [^01^] 
I fixed it by increasing the block number by 1, although my understanding is that the whole script happens within a single block so I'm not sure why it's like that.

---

The issue is present both in development and production script

[^01^]: ⚠ 77% possible duplicate - [Staking v2 production deploy](https://www.github.com/ubiquity/ubiquity-dollar/issues/995#995)

ryzhak:
Resolves https://github.com/ubiquity/ubiquity-dollar/issues/1001

Depends on https://github.com/ubiquity/ubiquity-dollar/pull/1003

This PR fixes staking migration which was broken after changes in https://github.com/ubiquity/ubiquity-dollar/pull/999

How to QA production staking: https://gist.github.com/ryzhak/78ffabebeef6540b4e085c132df1d6a8

Note, [here](https://github.com/ryzhak/ubiquity-dollar/blob/d224f73d32e08a0be53cc8397df3daffd1487775/packages/contracts/migrations/mainnet/Deploy002_Staking.s.sol#L76) we set `block.number + 10` because foundry works pretty weird (at least the latest v1.3.5.). It seems it caches `block.number` at the beginning of the script and later (if there was a new block mined) [here](https://github.com/ryzhak/ubiquity-dollar/blob/d224f73d32e08a0be53cc8397df3daffd1487775/packages/contracts/migrations/mainnet/Deploy002_Staking.s.sol#L76) doesn't fetch the latest `block.number` but uses the cached one. Hence the error `Error: reverted with: Can't start in the past` occurs. Setting it to `block.number + 10` helps both for QA and production deploy in case a new block is mined somewhere in the middle of migration execution. Setting staking start time to `block.number + 10` simply delays staking start time for ~2 mins.

### DIFF
--- a/packages/contracts/migrations/development/Deploy002_Staking.s.sol
+++ b/packages/contracts/migrations/development/Deploy002_Staking.s.sol
@@ -98,11 +98,10 @@ contract Deploy002_Staking is Script, DiamondTestHelper {
         stakingFacet.setGovernancePerBlock(0.2 ether); // 0.2 reward token minted per block
         stakingFacet.setGovernanceTreasuryDivider(5); // `100 / 5 = 20%` extra reward tokens minted for treasury
         stakingFacet.setStakingRewardToken(address(rewardToken)); // reward token address
-        stakingFacet.setStakingStartBlock(block.number); // activate staking from current block
+        stakingFacet.setStakingStartBlock(block.number + 10); // activate staking 10 blocks later
         stakingFacet.createStakingPool( // add a new staking pool
             100, // allocation points
-            IERC20(stakeToken), // staking token
-            true // whether to update all pools
+            IERC20(stakeToken) // staking token
         );
 
         // stop sending admin transactions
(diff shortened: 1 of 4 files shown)


whilefoo (commented):
shouldn't development script be also fixed?

ryzhak:
> shouldn't development script be also fixed?

Fixed in https://github.com/ubiquity/ubiquity-dollar/pull/1004/commits/c09cf213d867109a1cc46f4c8e4ebdad77102857

whilefoo (approved):

